PR #15, opened 2024-03-25
DR2-1539 Refactor monorepo.

MancunianSam:
The details of how this works are in the new README.

PR #19, opened 2024-03-28
DR2 1533 multiple preservation representations

MancunianSam:
- **DR2-1535 Update asset opex creator**
- **DR2-1534 Update parsed court document handler**
- **DR2-1536 Update reconciler**

Needs https://github.com/nationalarchives/dr2-ingest/pull/15 to be merged first


ingest-asset-reconciler/src/main/scala/uk/gov/nationalarchives/Lambda.scala
```diff
@@ -60,62 +60,73 @@ class Lambda extends LambdaRunner[Input, StateOutput, Config, Dependencies] {
...
-          IO.pure(
-            StateOutput(wasReconciled = false, s"There were no Content Objects returned for entity ref '${entity.ref}'")
-          )
-        else {
+      children <- childrenOfAsset(dependencies.dynamoDbClient, asset, config.dynamoTableName, config.dynamoGsiName)
+      _ <- IO.fromOption(children.headOption)(
+        new Exception(s"No children were found for ${input.assetId} from ${input.batchId}")
+      )
+      _ <- log(s"${children.length} children found for asset ${asset.id}")
+      groupedChildren = children.groupBy(_.representationType match {
```
techncl, line 68: 1. Shouldn't we separate the representation type groups also, something like

```
      groupedChildren =
        children.groupBy (
          child =>
            (child.representationType, child.representationSuffix) match {
              case (DynamoFormatters.PreservationRepresentationType, suffix) => (Preservation, suffix)
              case (DynamoFormatters.AccessRepresentationType, suffix)       => (Access, suffix)
            }
        )

      stateOutputs <- groupedChildren
        .map { case ((representationType, suffix), childrenForRepresentationType) =>
```

2. Maybe `groupedChildren` could be `childrenGroupedByRepType`?

ingest-asset-reconciler/src/main/scala/uk/gov/nationalarchives/Lambda.scala
```diff
@@ -60,62 +60,73 @@ class Lambda extends LambdaRunner[Input, StateOutput, Config, Dependencies] {
...
+
+                      bitstreamWithSameChecksum.isEmpty
+                    }
+
+                  if (childrenThatDidNotMatchOnChecksum.isEmpty) StateOutput(wasReconciled = true, "")
+                  else {
+                    val idsOfChildrenThatDidNotMatchOnChecksum = childrenThatDidNotMatchOnChecksum.map(_.id)
+                    StateOutput(
+                      wasReconciled = false,
+                      s"Out of the ${childrenForRepresentationType.length} files expected to be ingested for assetId '${input.assetId}' with representationType $representationType, " +
```
techncl, line 118: If we do what I suggested above, then maybe we should add the suffix to this message too?

ingest-parsed-court-document-event-handler/src/main/scala/uk/gov/nationalarchives/FileProcessor.scala
```diff
@@ -369,6 +375,10 @@ object FileProcessor {
     case File          => Json.fromString("File")
   }
 
+  sealed trait RepresentationType
```
techncl, line 378: Can't we import this from the Preservica client instead?

ingest-asset-reconciler/src/test/scala/uk/gov/nationalarchives/testUtils/ExternalServicesTestUtils.scala
```diff
@@ -84,6 +85,12 @@ class ExternalServicesTestUtils(dynamoServer: WireMockServer) extends TableDrive
        |      "digitalAssetSubtype": {
        |        "S": "Test Digital Asset Subtype"
        |      },
+       |      "representationType": {
+       |        "S": "Access"
+       |      },
+       |      "representationSuffix": {
+       |        "N": "2"
```
techncl, line 92: Even though it doesn't really matter since it's a test, shouldn't it be "1" since we start at 1 for each RepresentationType?

techncl (commented):
Looks good. Just a few comments

ingest-asset-opex-creator/src/test/scala/uk/gov/nationalarchives/XMLCreatorTest.scala
```diff
@@ -70,9 +75,16 @@ class XMLCreatorTest extends AnyFlatSpec {
...
-        <Name>Preservation</Name>
+        <Name>Preservation_1</Name>
         <ContentObjects>
           <ContentObject>a814ee41-89f4-4975-8f92-303553fe9a02</ContentObject>
+        </ContentObjects>
+      </Representation>
+      <Representation>
+        <InformationObject>90730c77-8faa-4dbf-b20d-bba1046dac87</InformationObject>
+        <Type>Access</Type>
+        <Name>Access_2</Name>
```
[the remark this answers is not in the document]
MancunianSam: Yeah, fair point. I think I was trying to show it writes the extra suffix but it makes more sense to have each as _1 I think.

ingest-asset-reconciler/src/main/scala/uk/gov/nationalarchives/Lambda.scala
MancunianSam: 1. I've tried this and it's not giving a `RepresentationType` as the key type of the Map because it can't tell that the match for `(child.representationType, child.representationSuffix)` is exhaustive (I think)
Aside from that, I don't think it's needed, the existing code and this are equivalent. 
2. Yeah, ok.

MancunianSam: I'm getting errors trying this further down. I could fix them but it's a fair bit of extra code and a type signature of `Map[(EntityClient.RepresentationType, Int), List[FileDynamoTable]]` for not much benefit so I'll leave it I think.
I've changed the variable name.

ingest-asset-opex-creator/src/test/scala/uk/gov/nationalarchives/XMLCreatorTest.scala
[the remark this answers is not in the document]
techncl: I guess we could also add a `Preservation_2` or `Access_2` to the XML to show that we can do that?

ingest-asset-reconciler/src/main/scala/uk/gov/nationalarchives/Lambda.scala
MancunianSam: As above, the error message as is is helpful enough not to need the extra code.

ingest-parsed-court-document-event-handler/src/main/scala/uk/gov/nationalarchives/FileProcessor.scala
techncl: Cool. I was also thinking about having common case classes together.

techncl (approved):
👍

MancunianSam:
@techncl Can you re-approve. I've also pushed a minor change to the github workflows as we were just deploying the same version over and over.

techncl (approved):
👍

PR #29, opened 2024-04-05
Write the suffix as an integer.

MancunianSam:
We're expecting this to be an integer in Dynamo but it's a string in the
Bagit json. This should write it to the bagit file as a number so it
should end up in Dynamo as a number.


techncl (approved):
👍

PR #42, opened 2024-04-10
Account for dots or lack of extensions in titles

techncl:


MancunianSam (approved):
Thanks for sorting this.

PR #49, opened 2024-04-17
Dr2 1511 check preservica for an existing Information Object

techncl:


ingest-check-preservica-for-existing-io/src/main/scala/uk/gov/nationalarchives/Lambda.scala
```diff
@@ -0,0 +1,84 @@
...
+
+      entitiesWithAssetName <- dependencies.entityClient.entitiesByIdentifier(Identifier(sourceId, asset.name))
+
+      potentialIoExistsResponse = entitiesWithAssetName.headOption match {
+        case Some(entityWithAssetName) =>
+          val potentialEntityType: Option[EntityType] = entityWithAssetName.entityType
+          if (potentialEntityType.contains(InformationObject)) Right(true)
+          else
+            Left(
+              new Exception(
```
MancunianSam, line 50: Do you need to throw an exception if it's not an InformationObject? Is it possible that anything but an IO will have a SourceId attached or is it possible in the future that we might legitimately want to put a SourceId on another entityType? I'm not sure but it feels like the purpose of this lambda is to check if there is an IO with that source ID. If it's not an IO, it returns false. 

The other thing is that we do sometimes get odd results from Preservica so being as permissive as possible is usually good.

Also, we've got tests on the opex creator which should check that we're putting the SourceId in the right place. 

What do you think? You could simplify this a fair bit as well because it would just become:

```scala
assetExists = entitiesWithAssetName.headOption.contains(InformationObject)
```

ingest-check-preservica-for-existing-io/src/main/resources/application.conf
```diff
@@ -0,0 +1,3 @@
+api-url = ${PRESERVICA_API_URL}
+secret-name = ${PRESERVICA_SECRET_NAME}
+dynamo-table-name = ${DYNAMO_TABLE_NAME}
```
MancunianSam, line 3: Can you add a line break please.

MancunianSam (commented):
Looks good. Just a couple of bits. You don't need the application.conf file in the test package either.

ingest-check-preservica-for-existing-io/src/main/scala/uk/gov/nationalarchives/Lambda.scala
techncl: Ok. yeah, this was a last minute addition. I was thinking thinking in case we accidentally/purposely add a `sourceId` to a folder or file. I'll change it back

MancunianSam (approved):
:+1:

PR #50, opened 2024-04-18
DR2-1590 Change the workflow to use the new name

MancunianSam:
They're all prefixed with {environment}-dr2- so this should work.


techncl (dismissed):
👍

techncl (approved):

PR #62, opened 2024-04-29
Use a mock class instead of Wiremock for the DiscoveryService

techncl:


MancunianSam (approved):
This is loads better, thanks for sorting it out.

PR #78, opened 2024-05-08
Get message from lock table and then return fields in Reconciler

techncl:


ingest-asset-reconciler/src/main/scala/uk/gov/nationalarchives/Lambda.scala
```diff
@@ -39,34 +42,107 @@ class Lambda extends LambdaRunner[Input, StateOutput, Config, Dependencies] {
...
+      childrenForRepresentationType.filter { assetChild =>
+        val bitstreamWithSameChecksum = bitstreamInfoPerContentObject.find { bitstreamInfoForCo =>
+          assetChild.checksumSha256 == bitstreamInfoForCo.fixity.value &&
+          coTitleMatchesAssetChildTitle(bitstreamInfoForCo.potentialCoTitle, assetChild)
+        }
+
+        bitstreamWithSameChecksum.isEmpty
+      }
+
+    lazy val idsOfChildrenThatDidNotMatchOnChecksum = childrenThatDidNotMatchOnChecksum.map(_.id)
```
MancunianSam, line 80: Howcome this is `lazy val`? Also, do you need this interim variable at all, it's only used once in the output and you could put `childrenThatDidNotMatchOnChecksum.map(_.id)` straight in the string. 
The variable name is very long and it might tidy it up a bit, what do you think?

ingest-asset-reconciler/src/main/scala/uk/gov/nationalarchives/Lambda.scala
```diff
@@ -39,34 +42,107 @@ class Lambda extends LambdaRunner[Input, StateOutput, Config, Dependencies] {
...
+      items <- dependencies.dynamoDbClient.getItems[IngestLockTable, PartitionKey](
+        List(PartitionKey(assetId)),
+        lockTableName
+      )
+
+      attributes <- IO.fromOption(items.headOption)(
+        new Exception(s"No items found for assetId '$assetId' from batchId '$batchId'")
+      )
+
+      message <- IO.fromEither(decode[AssetMessage](attributes.message.replace('\'', '\"')))
```
MancunianSam, line 102: What's the call to `replace('\''...` for?

MancunianSam (commented):
Looks good, just a couple of questions.

ingest-asset-reconciler/src/main/scala/uk/gov/nationalarchives/Lambda.scala
techncl: With this code, I moved what was already there, to methods but also decided to do a minor clean up

1 . This line was initially in the `else` block, so I moved it here so that I could remove the braces from the `else` to make it cleaner; it's a `lazy val` so that it's not run in the case that `childrenThatDidNotMatchOnChecksum.isEmpty` i.e. it only runs the code when used
2. I'm not sure why I saved it to a variable, maybe it was used more than once initially but yes, it's a better idea to add it directly to the string and then we no longer have the long variable name

ingest-asset-reconciler/src/main/scala/uk/gov/nationalarchives/Lambda.scala
techncl: The tests kept throwing this error `software.amazon.awssdk.thirdparty.jackson.core.JsonParseException: Unexpected character ('p' (code 112)): was expecting comma to separate Object entries`; it's got to do with the lock-table's mock response JSON. It has an issue with me using double quotes in the `message` (the `message` itself is a JSON, so json within a json) so I replaced the double quotes in the JSON `message` with single quotes and then did a `.replace` reversal right before the decoding as this was the only place I could add it where it wouldn't get the error.

PR #87, opened 2024-05-21
Refactor ingest mapper

techncl:


MancunianSam (approved):
Nice, thanks for tidying this up.

PR #97, opened 2024-05-30
Move assetName out of SNS message and into main output & reaplace with assetId

techncl:
'assetName' is not needed for SNS message but it is needed for the next step in step function, therefore, it makes more sense to put it in the main output.

In the SNS message, assetId was replaced with assetName but we actually need it for the message


MancunianSam (approved):
:+1:

PR #99, opened 2024-06-03
Correct the by-identifier endpoint.

MancunianSam:
The endpoint is /entity/entities/by-identifier. We were only calling
/entity/by-identifier.


techncl (approved):
🤦 . Thanks!

PR #101, opened 2024-06-05
Make reconciliation failure messages more prominent

techncl:


kurtismash:
Would this change be better placed elsewhere in our ingest process? On our team day we added a ticket to the backlog to ["check reconciler actions - is it pure?"](https://national-archives.atlassian.net/browse/DR2-1666), this Lambda function is taking on a lot of additional tasks to generate message metadata and now presentation of logs.

techncl:
Oh, I thought we decided yesterday that it we were going to put it in the Reconciler Lambda?

MancunianSam:
I think considering this is already here and it's a fairly quick win to get some better messages, this is fine. When we move the notifications, we can just lift this change straight into its new home.

MancunianSam (approved):

PR #103, opened 2024-06-07
DR2-1700 Rename packages.

MancunianSam:
This includes a Lambda class in the old uk.gov.nationalarchives
directory that extends the new one. This should allow us to release this
without breaking things.

I've also renamed the ingest-check-preservica-for-existing-io in
preparation for DR2-1654 but kept the lambda name so I don't break that.


techncl (dismissed):
Looks good!

techncl (approved):

PR #108, opened 2024-06-10
DR2-1690 Move Dynamo Formatters

MancunianSam:
This is now another module in the project.

### DIFF
--- a/ingest-upsert-archive-folders/src/test/scala/uk/gov/nationalarchives/ingestupsertarchivefolders/testUtils/ExternalServicesTestUtils.scala
+++ b/ingest-upsert-archive-folders/src/test/scala/uk/gov/nationalarchives/ingestupsertarchivefolders/testUtils/ExternalServicesTestUtils.scala
@@ -14,13 +14,13 @@ import org.scalatestplus.mockito.MockitoSugar
 import org.scanamo.DynamoFormat
 import software.amazon.awssdk.services.eventbridge.model.PutEventsResponse
 import sttp.capabilities.fs2.Fs2Streams
-import uk.gov.nationalarchives.dynamoformatters.DynamoFormatters.*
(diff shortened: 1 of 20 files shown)


ingest-upsert-archive-folders/src/test/scala/uk/gov/nationalarchives/ingestupsertarchivefolders/testUtils/ExternalServicesTestUtils.scala
```diff
@@ -31,6 +32,8 @@ import scala.jdk.CollectionConverters.*
 
 class ExternalServicesTestUtils extends AnyFlatSpec with BeforeAndAfterEach with BeforeAndAfterAll with TableDrivenPropertyChecks with MockitoSugar {
 
+  case class TestIdentifier(name: String, value: String)
```
techncl, line 35: How come we added this?

techncl (commented):
Looks good! Just 1 question, as well as thes:, 

1. Are `DynamoIdentifier` and `PreservicaIdentifier` the same thing i.e. does one eventually become the other? if so, could we just have one in the Preservica client so that we can ensure that they are consistent?
2. Do you think that maybe `EntityIdentifier` might be a clearer name than `PreservicaIdentifier`? Or maybe in the Preservica client, we could rename `Identifier` to `EntityIdentifier`?

ingest-upsert-archive-folders/src/test/scala/uk/gov/nationalarchives/ingestupsertarchivefolders/testUtils/ExternalServicesTestUtils.scala
MancunianSam: The `Table` we're using for the tests has a mix of Preservica `Identifier` and Dynamo `Identifier`. This was fine when they were the same thing but now it's got a bit messy. I couldn't get it to compile in any kind of neat way so I introduced this as a sort of "bridge" between the two for testing purposes.

techncl (approved):
👍

PR #109, opened 2024-06-10
DR2-1702 Remove temporary classes.

MancunianSam:
These were added so we could deploy without breaking things. The
terraform changes are in so these can be deleted.


techncl (approved):
👍

PR #114, opened 2024-06-11
DR2-1716 Update lambda name

MancunianSam:
This is a case of removing the override that gave it the old name. This
can't be deployed until
https://github.com/nationalarchives/dr2-terraform-environments/pull/124
is merged and deployed.


techncl (approved):
👍

PR #116, opened 2024-06-12
DR2-1709 Add child count to case classes

MancunianSam:
DR2-1709 Add childCount to Dynamo case classes

This allows us to read the childCount out of Dynamo. The ticket says that this should be on all rows so I've made it mandatory.

### DIFF
--- a/ingest-asset-opex-creator/src/test/scala/uk/gov/nationalarchives/ingestassetopexcreator/testUtils/ExternalServicesTestUtils.scala
+++ b/ingest-asset-opex-creator/src/test/scala/uk/gov/nationalarchives/ingestassetopexcreator/testUtils/ExternalServicesTestUtils.scala
@@ -155,6 +155,9 @@ class ExternalServicesTestUtils(dynamoServer: WireMockServer, s3Server: WireMock
        |      "sortOrder": {
        |        "N": "1"
        |      },
+       |      "childCount": {
+       |        "N": "0"
+       |      },
        |      "id": {
        |        "S": "$childIdDocx"
        |      },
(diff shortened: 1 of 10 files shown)


ingest-asset-opex-creator/src/test/scala/uk/gov/nationalarchives/ingestassetopexcreator/testUtils/ExternalServicesTestUtils.scala
```diff
@@ -155,6 +155,9 @@ class ExternalServicesTestUtils(dynamoServer: WireMockServer, s3Server: WireMock
        |      "sortOrder": {
        |        "N": "1"
        |      },
+       |      "childCount": {
+       |        "N": "0"
```
techncl, line 159: Just out of curiosity, is it possible to have 0 children? If not, should we do a check/prevent this from happening?

techncl (commented):
Looks good! Just one question as well as this

Since `childCount` is a numeric field, do you think we should add the `invalidNumericField` and `invalidNumericValue` tests for it?

ingest-asset-opex-creator/src/test/scala/uk/gov/nationalarchives/ingestassetopexcreator/testUtils/ExternalServicesTestUtils.scala
MancunianSam: It's possible to have zero children. Files don't have any children.

MancunianSam:
Good point about the `invalidNumericField` and `invalidNumericValue` tests. I've added them in.

PR #119, opened 2024-06-12
DR2-1692 Asset opex creator changes.

MancunianSam:
This throws an exception if the childCount doesn't equal the number of
children. I've added in a test to check for it.


ingest-asset-opex-creator/src/main/scala/uk/gov/nationalarchives/ingestassetopexcreator/Lambda.scala
```diff
@@ -38,6 +38,7 @@ class Lambda extends LambdaRunner[Input, Unit, Config, Dependencies] {
       _ <- log(s"Asset ${asset.id} retrieved from Dynamo")
 
       children <- childrenOfAsset(dynamoClient, asset, config.dynamoTableName, config.dynamoGsiName)
+      _ <- IO.raiseWhen(children.length != asset.childCount)(new Exception(s"Asset ${asset.id}: Expected ${asset.childCount} children but ${children.length} children found"))
```
techncl, line 41: Even though what this error means is obvious, I was thinking about future us/employees reading it:

Do you think we should add add where the counts came from? e.g.
s"Asset ${asset.id}: Expected ${asset.childCount} children but found ${children.length} children in the Preservation system"
Or

s"Asset ${asset.id}: has ${asset.childCount} children in the files table but found ${children.length} children in the Preservation system"?

Maybe "Asset id ${asset.id}" or "'assetId' ${asset.id}" instead of "Asset ${asset.id}" just because the id and name are both uuids

techncl (commented):
Looks good! Just a minor comment, as well as this:

Should we account for the `childCount` attribute being absent?

ingest-asset-opex-creator/src/main/scala/uk/gov/nationalarchives/ingestassetopexcreator/Lambda.scala
MancunianSam: Good point, I've gone with `s"Asset id ${asset.id}: has ${asset.childCount} children in the files table but found ${children.length} children in the Preservation system"`

techncl (approved):
👍

PR #120, opened 2024-06-12
DR2-1693 Folder opex creator changes.

MancunianSam:
This checks the childCount against the returned number of children.
There's a unit test for it as well.


ingest-folder-opex-creator/src/main/scala/uk/gov/nationalarchives/ingestfolderopexcreator/Lambda.scala
```diff
@@ -87,6 +87,7 @@ class Lambda extends LambdaRunner[Input, Unit, Config, Dependencies] {
       _ <- log(s"Fetched ${folderItems.length} folder items from Dynamo")
 
       children <- childrenOfFolder(dependencies.dynamoClient, folder, config.dynamoTableName, config.dynamoGsiName)
+      _ <- IO.raiseWhen(folder.childCount != children.length)(new Exception(s"Folder ${folder.id}: Expected ${folder.childCount} children but found ${children.length} children"))
```
techncl, line 90: Even though what this error means is obvious, I was thinking about future us/employees reading it:

Do you think we should add add where the counts came from? e.g.
s"Folder ${folder.id}: Expected ${folder.childCount} children but found ${children.length} children in the Preservation system"

Or

s"Folder ${folder.id}: has ${folder.id} children in the files table but found ${children.length} children in the Preservation system"?

Maybe "Folder id ${folder.id}" instead of "Folder ${folder.id}"?

techncl (commented):
Looks good! Just a minor comment, as well as this:

Should we account for the `childCount` attribute being absent?

ingest-folder-opex-creator/src/main/scala/uk/gov/nationalarchives/ingestfolderopexcreator/Lambda.scala
MancunianSam: I've gone with `s"Folder id ${folder.id}: has ${folder.childCount} children in the files table but found ${children.length} children in the Preservation system"`

MancunianSam:
As I said in the other one, childCount is mandatory so this will error if it's missing.

techncl (approved):
👍

PR #121, opened 2024-06-12
DR2-1694 Reconciler changes.

MancunianSam:


ingest-asset-reconciler/src/main/scala/uk/gov/nationalarchives/ingestassetreconciler/Lambda.scala
```diff
@@ -147,6 +147,7 @@ class Lambda extends LambdaRunner[Input, StateOutput, Config, Dependencies] {
       )
 
       children <- childrenOfAsset(dependencies.dynamoDbClient, asset, config.dynamoTableName, config.dynamoGsiName)
+      _ <- IO.raiseWhen(asset.childCount != children.length)(new Exception(s"Asset $assetId: Expected ${asset.childCount} children but found ${children.length} children"))
```
techncl, line 150: Even though what this error means is obvious, I was thinking about future us/employees reading it:

1. Do you think we should add add where the counts came from? e.g.
`s"Asset $assetId: Expected ${asset.childCount} children but found ${children.length} children in the Preservation system"`

 Or

`s"Asset $assetId: has ${asset.childCount} children in the files table but found ${children.length} children in the Preservation system"`?

2. Maybe "Asset id $assetId" or "'assetId' $assetId" instead of "Asset $assetId" just because the id and name are both uuids

techncl (approved):

ingest-asset-reconciler/src/main/scala/uk/gov/nationalarchives/ingestassetreconciler/Lambda.scala
MancunianSam: I've changed it to the same as the asset opex creator.

techncl (approved):
👍

PR #142, opened 2024-06-27
DR2-1729 Change primary key

MancunianSam:
This has been changed in the Dynamo formatters and updated everywhere
else.


ingest-find-existing-asset/src/test/scala/uk/gov/nationalarchives/ingestfindexistingasset/testUtils/ExternalServicesTestUtils.scala
```diff
@@ -175,6 +175,8 @@ class ExternalServicesTestUtils extends AnyFlatSpec with BeforeAndAfterEach with
       val serveEvents = dynamoServer.getAllServeEvents.asScala.toList
 
       serveEvents.size should equal(numOfDynamoGetRequests + numOfDynamoUpdateRequests)
+      val d = serveEvents.map(_.getRequest.getBodyAsString)
```
techncl, line 178: I'm guessing these two lines were left in by mistake?

dynamo-formatters/src/main/scala/uk/gov/nationalarchives/dynamoformatters/DynamoFormatters.scala
```diff
@@ -199,7 +199,7 @@ object DynamoFormatters {
 
   case class Identifier(identifierName: String, value: String)
 
-  case class FilesTablePartitionKey(id: UUID)
+  case class FilesTablePartitionKey(id: UUID, batchId: String)
```
techncl, line 202: 1. Should this be renamed to `FilesTablePrimaryKey`?
2. I could be wrong, but from what I understand, the `batchId` will be the "sort key", right? I guess for clarity/correctness, should we do this :

```
case class FilesTablePartitionKey(id: UUID)
case class FilesTableSortKey(batchId: String)

case class FilesTablePrimaryKey(FilesTablePartitionKey, FilesTableSortKey)

```

techncl (commented):
Looks good! Just two comments

dynamo-formatters/src/main/scala/uk/gov/nationalarchives/dynamoformatters/DynamoFormatters.scala
MancunianSam: 1. Ok. I've renamed the class. 
2. That makes sense. I've changed it to `case class FilesTablePrimaryKey(partitionKey: FilesTablePartitionKey, sortKey: FilesTableSortKey)` and updated the `DynamoFormatter` so that it reads and writes it in the `{"batchId":{"S":"TEST-ID"},"id":{"S":"assetId"}}`

ingest-find-existing-asset/src/test/scala/uk/gov/nationalarchives/ingestfindexistingasset/testUtils/ExternalServicesTestUtils.scala
MancunianSam: Ah yeah, I was debugging. I've removed them.

ingest-upsert-archive-folders/src/main/scala/uk/gov/nationalarchives/ingestupsertarchivefolders/LambdaRunner.scala
```diff
@@ -0,0 +1,41 @@
...
+
+import cats.effect.IO
+import uk.gov.nationalarchives.{DADynamoDBClient, DAEventBridgeClient}
+import uk.gov.nationalarchives.dp.client.fs2.Fs2Client
+import uk.gov.nationalarchives.ingestupsertarchivefolders.Lambda.{Config, Dependencies, StepFnInput}
+
+import cats.effect.unsafe.implicits.global
+
+object LambdaRunner extends App {
+  """
```
techncl, line 11: 1. This file belongs to the upsert Lambda; did you edit it deliberately?
2. This string isn't saved to a variable

techncl (approved):
👍

PR #146, opened 2024-07-01
Dr2 1706 add ttl attribute to all items in ingest mapper

techncl:


ingest-mapper/src/main/scala/uk/gov/nationalarchives/ingestmapper/Lambda.scala
```diff
@@ -56,12 +59,22 @@ class Lambda extends LambdaRunner[Input, StateOutput, Config, Dependencies] {
...
 
       bagManifests <- dependencies.metadataService.parseBagManifest(input)
       bagInfoJson <- dependencies.metadataService.parseBagInfoJson(input)
-      metadataJson <- dependencies.metadataService.parseMetadataJson(input, departmentAndSeries, bagManifests, bagInfoJson.headOption.getOrElse(Obj()))
+      metadataJson <- dependencies.metadataService.parseMetadataJson(
+        input,
+        departmentAndSeries,
+        bagManifests,
+        bagInfoJson.headOption.getOrElse(Obj()),
+        hundredDaysFromNowInEpochSecs
```
MancunianSam, line 76: Instead of having to pass what is essentially a constant into two different methods, would it be easier to add the ttl attribute in the line below.
Something like
```scala
metadataJsonWithTtl <- IO(metadataJson.map(obj => Obj.from(obj.value ++ Map("ttl" -> hundredDaysFromNowInEpochSecs))))
```
This way, you guarantee you've got all the items as it's just before the write to dynamo and you don't have to pass around the hundred years parameter. 
It also means you're not testing essentially the same thing in each test class.

You'd have to re-jig the tests a bit, I got it running on a branch though.  https://github.com/nationalarchives/dr2-ingest/compare/DR2-1706_addTtlAttributeToAllItemsInIngestMapper...suggested-changes?expand=1

ingest-mapper/src/test/scala/uk/gov/nationalarchives/ingestmapper/testUtils/LambdaTestTestUtils.scala
```diff
@@ -104,7 +105,7 @@ class LambdaTestTestUtils(dynamoServer: WireMockServer, s3Server: WireMockServer
     }
   }
 
-  def checkDynamoItems(tableRequestItems: List[DynamoTableItem], expectedTable: DynamoTable): Assertion = {
+  def checkDynamoItems(tableRequestItems: List[DynamoTableItem], expectedTable: DynamoFilesTableItem): Assertion = {
     val items = tableRequestItems
```
MancunianSam, line 109: You're not checking for the ttl entry in this method.

MancunianSam (commented):
Looks good. I think we can simplify it a bit though, what do you think?

MancunianSam (approved):
:+1:

PR #175, opened 2024-07-24
DR2-1793 remove user and secret env vars

MancunianSam:
This now gets them from the secret

There is a point where we're calling `head` on the map to get the secret
username and password. I'm usually against calling `head` without checks
but in this case there can be only one entry in secrets manager so we're
ok.

### DIFF
--- a/rotate-preservation-system-password/src/main/scala/uk/gov/nationalarchives/rotatepreservationsystempassword/Lambda.scala
+++ b/rotate-preservation-system-password/src/main/scala/uk/gov/nationalarchives/rotatepreservationsystempassword/Lambda.scala
@@ -19,13 +19,13 @@ import scala.jdk.CollectionConverters.*
 
 class Lambda extends LambdaRunner[RotationEvent, Unit, Config, Dependencies] {
 
-  override def dependencies(config: Config): IO[Dependencies] = userClient(config.apiUrl, config.secretName).map { userClient =>
-    Dependencies(userClient, secretId => DASecretsManagerClient[IO](secretId))
-  }
+  override def dependencies(config: Config): IO[Dependencies] =
+    IO.pure(Dependencies(secretId => userClient(config.apiUrl, secretId), secretId => DASecretsManagerClient[IO](secretId)))
 
   override def handler: (RotationEvent, Config, Dependencies) => IO[Unit] = (event, config, dependencies) =>
     for {
       client <- IO.pure(dependencies.secretsManagerClient(event.secretId))
+      userClient <- dependencies.userClient(event.secretId)
       token = event.clientRequestToken
       secretId = event.secretId
       describeSecretResponse <- client.describeSecret()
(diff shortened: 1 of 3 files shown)


rotate-preservation-system-password/src/main/scala/uk/gov/nationalarchives/rotatepreservationsystempassword/Lambda.scala
```diff
@@ -19,13 +19,13 @@ import scala.jdk.CollectionConverters.*
 
 class Lambda extends LambdaRunner[RotationEvent, Unit, Config, Dependencies] {
 
-  override def dependencies(config: Config): IO[Dependencies] = userClient(config.apiUrl, config.secretName).map { userClient =>
-    Dependencies(userClient, secretId => DASecretsManagerClient[IO](secretId))
-  }
+  override def dependencies(config: Config): IO[Dependencies] =
+    IO.pure(Dependencies(secretId => userClient(config.apiUrl, secretId), secretId => DASecretsManagerClient[IO](secretId)))
```
techncl, line 23: Shouldn't this be `IO` rather than `IO.pure` since it calls the `userClient`?

rotate-preservation-system-password/src/main/scala/uk/gov/nationalarchives/rotatepreservationsystempassword/Lambda.scala
```diff
@@ -35,12 +35,16 @@ class Lambda extends LambdaRunner[RotationEvent, Unit, Config, Dependencies] {
...
-        case SetSecret    => setSecret(client, dependencies.userClient, config.apiUser)
-        case TestSecret   => dependencies.userClient.testNewPassword()
+        case CreateSecret => createSecret(client, token)
+        case SetSecret    => setSecret(client, userClient)
+        case TestSecret   => userClient.testNewPassword()
         case FinishSecret => finishSecret(client, token)
     } yield ()
 
+  private def getAuthDetailsFromSecret(client: DASecretsManagerClient[IO], stage: Stage) = for {
+    currentSecret <- client.getSecretValue[Map[String, String]](stage)
```
techncl, line 45: What do you think about this:

```
  private def getAuthDetailsFromSecret(client: DASecretsManagerClient[IO], stage: Stage) = for {
    (username, password) <- client.getSecretValue[Map[String, String]](stage).map(_.head)
  } yield AuthDetails(username, password)
```

techncl (commented):
Looks good! Just 2 comments

rotate-preservation-system-password/src/main/scala/uk/gov/nationalarchives/rotatepreservationsystempassword/Lambda.scala
MancunianSam: That's better, gets rid of `._1` and `._2` thanks.

techncl (approved):
👍

PR #178, opened 2024-07-26
DR2-1713 Remove notification logic

MancunianSam:


techncl (commented):
We're removing this but where are we generating this notification now?

MancunianSam:
It's going to be the files change handler that generates these now.

techncl (approved):
👍

PR #194, opened 2024-08-05
DR2-1841 Restructure repo

MancunianSam:
We're getting a few Python lambdas now and the custodial copy stuff
could do with coming into the monorepo. This is in preperation for that.


build.sbt
```diff
@@ -1,250 +1,9 @@
...
-  .settings(commonSettings)
-  .dependsOn(utils)
+  .aggregate(ingestLambdasRoot)
   .settings(
-    libraryDependencies ++= Seq(
-      preservicaClient,
-      secretsManagerClient
-    )
+    name := "preservation",
+    scalaVersion := "3.4.2"
```
techncl, line 5: 1. There is a Scala version in the `build.sbt` file; do we need to have them in both places?
2. How come the `name` here is "preservation" and in the `build.sbt` it's "lambdas"?

techncl (commented):

build.sbt
MancunianSam: 1. No, we don't need them in both places, thanks. It just needs to be declared once in the root build.sbt
2. The file project with the name "preservation" will be an aggregate of the lambdas subproject and eventually the docker images (custodial copy) stuff as well. So it will be something like
```
.
├── build.sbt
├── python
│   └── lambdas
│       └── copy-files-from-tdr
└── scala
    ├── images
    │   └── build.sbt
    └── lambdas
        ├── build.sbt
        ├── dynamo-formatters
        ├── entity-event-generator-lambda
        ├── get-latest-preservica-version-lambda
        ├── ingest-asset-opex-creator
        ├── ingest-asset-reconciler
        ├── ingest-files-change-handler
        ├── ingest-find-existing-asset
         ....
```
So the name for the `images/build.sbt` will be dockerImages or something. I want to group things by how they're deployed. So the name in the root file needs to encompass everything we've got so I went with preservation.

techncl: Ah, I see. Would "ingest" work or are we expecting to add non-ingest stuff here?

MancunianSam: No, we're not, ingest makes more sense. I'll change it.

techncl (approved):
👍 . There's an trailing comma; not sure if that was on purpose

PR #197, opened 2024-08-08
Minor lambda changes.

MancunianSam:
The parsed court document handler was loading its own config rather than
using the config passed to it. I've updated this.

The files change handler is throwing an error if there is an unexpected
number of children. We talked about this a little while ago and this can
be valid. At any rate, these errors are making it difficult to see if
the process is working so I've taken them out.
I've also added a couple of logging lines, I'll remove them later.


techncl (approved):
👍

PR #225, opened 2024-09-06
Fix concurrency issues.

MancunianSam:
The problem we're having is that each fiber was running in parallel and
trying to get the Group from the Map at the same time. Ref allows
parallel access to the Map. All the fibers were finding the Map empty
and starting a new group.

The solution is to use AtomicCell which only allows one thread to access
it at once. To make this work, I've refactored it so the read and write
of the map are inside the same evalMap call on AtomicCell.

### DIFF
--- a/scala/lambdas/preingest-tdr-aggregator/src/main/scala/uk/gov/nationalarchives/preingesttdraggregator/Aggregator.scala
+++ b/scala/lambdas/preingest-tdr-aggregator/src/main/scala/uk/gov/nationalarchives/preingesttdraggregator/Aggregator.scala
@@ -76,15 +76,15 @@ object Aggregator:
       )
     }
 
-    private def startNewGroup(sourceId: String, config: Config, groupExpiryTime: Long)(using
+    private def startNewGroup(sourceId: String, config: Config, groupExpiryTime: MilliSeconds)(using
         sfnClient: DASFNClient[F],
         enc: Encoder[SFNArguments]
     ): F[Group] = {
-      val waitFor: Seconds = Math.ceil((groupExpiryTime - Generators().generateInstant.toEpochMilli).toDouble / 1000).toInt.seconds
+      val waitFor: Seconds = Math.ceil((groupExpiryTime - Generators().generateInstant.toEpochMilli.milliSeconds).toDouble / 1000).toInt.seconds
       val groupId: GroupId = GroupId(config.sourceSystem)
       val batchId = BatchId(groupId)
       sfnClient.startExecution(config.sfnArn, SFNArguments(groupId, batchId, waitFor), batchId.batchValue.some).map { _ =>
-        Group(groupId, Instant.ofEpochMilli(groupExpiryTime), 1)
+        Group(groupId, Instant.ofEpochMilli(groupExpiryTime.length), 1)
       }
     }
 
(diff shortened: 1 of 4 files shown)


scala/lambdas/preingest-tdr-aggregator/src/main/scala/uk/gov/nationalarchives/preingesttdraggregator/Aggregator.scala
```diff
@@ -81,45 +76,45 @@ object Aggregator:
...
-        val potentialCurrentGroupForSource = groupCache.get(sourceId)
-        potentialCurrentGroupForSource match
-          case None => log(NoExistingGroup) >> startNewGroup(groupCacheRef, sourceId, config, lambdaTimeoutTime)
-          case Some(currentGroupForSource) =>
-            if currentGroupForSource.expires.toEpochMilli <= lambdaTimeoutTime then
-              log(ExpiryBeforeLambdaTimeout) >> startNewGroup(groupCacheRef, sourceId, config, lambdaTimeoutTime)
-            else if currentGroupForSource.itemCount > config.maxBatchSize then log(MaxGroupSizeExceeded) >> startNewGroup(groupCacheRef, sourceId, config, lambdaTimeoutTime)
-            else groupCacheRef.update(_ + (sourceId -> currentGroupForSource.copy(itemCount = currentGroupForSource.itemCount + 1))) >> Async[F].pure(currentGroupForSource.groupId)
-      }
+      val groupExpiryTime = lambdaTimeoutTime + (config.maxSecondaryBatchingWindow * 1000)
```
techncl, line 95: Could this be converted to `Milliseconds`?

techncl (commented):
Looks good! Just one comment.

scala/lambdas/preingest-tdr-aggregator/src/main/scala/uk/gov/nationalarchives/preingesttdraggregator/Aggregator.scala
MancunianSam: Yeah, it can. I've changed a couple of bits so we've got Seconds and MilliSeconds in most places.

scala/lambdas/preingest-tdr-aggregator/src/main/scala/uk/gov/nationalarchives/preingesttdraggregator/Duration.scala
```diff
@@ -17,10 +17,12 @@ object Duration:
     def *(other: Int): Int = s * other
     def length: Int = s
     def toJson: Json = Json.fromInt(s)
+    def toMilliSeconds: MilliSeconds = s * 1000
```
techncl, line 20: I just realised that the S is capitalised; could you lowercase it?

techncl (commented):
Looks good! just one additional comment

scala/lambdas/preingest-tdr-aggregator/src/main/scala/uk/gov/nationalarchives/preingesttdraggregator/Duration.scala
MancunianSam: Ah yeah. I was thinking milli seconds as two words but it's not :facepalm: I've changed it.

PR #236, opened 2024-09-12
Add cc notification handler to monorepo

MancunianSam:
This is more or less a straight move from the existing repo with a few tweaks to get the tests working with the other python lambda.


python/lambdas/ingest-cc-notifications/LICENCE
```diff
@@ -0,0 +1,18 @@
+Copyright (C) 2024 Crown Copyright (The National Archives)
```
kurtismash, line 1: A general question, what happened the title in our license files? https://gds-way.digital.cabinet-office.gov.uk/manuals/licensing.html#use-mit

MancunianSam: I'm not sure what's happened here. There is a licence in the root of this project in the correct format though so we don't need this one.

techncl (approved):
👍

PR #237, opened 2024-09-12
DR2-1817 Add projection expression to query

MancunianSam:
Also tidy up the imports


kurtismash (approved):

python/lambdas/ingest-cc-notifications/cc_notifications/lambda_function.py
```diff
@@ -15,7 +11,8 @@ def get_items_with_id(client, table_name, primary_key, primary_key_value):
     response = client.query(
         TableName=table_name,
         KeyConditionExpression=f"{primary_key} = :value",
-        ExpressionAttributeValues={":value": {"S": primary_key_value}}
+        ExpressionAttributeValues={":value": {"S": primary_key_value}},
+        ProjectionExpression=f"id,batchId,{attribute_to_add}"
```
techncl, line 15: Should we update the tests to reflect that the `ingested_CC` is also returned?

MancunianSam: How do you mean? Because you're using moto, it behaves like a real Dynamo table so as long as the existing tests still pass, it shows that this change is working I think?

techncl (approved):

PR #241, opened 2024-09-13
DR2-1879 Get the correct checksum

MancunianSam:
I can't run the stream twice, once to upload it and once to get the
checksum. It might be possible to do it with streams but it's
complicated by the fact that we're parsing the tar file as well so I
give up.

This now compiles the stream into a byte array, uses the standard
apache commons DigestUtils to get the checksum and then creates a new
stream to allow us to upload the file.

It's inefficient but we're only ever going to be dealing with small
files for judgments.


scala/lambdas/ingest-parsed-court-document-event-handler/src/main/scala/uk/gov/nationalarchives/ingestparsedcourtdocumenteventhandler/FileProcessor.scala
```diff
@@ -174,22 +173,17 @@ class FileProcessor(
...
               Stream.eval[IO, (String, FileInfo)](
-                stream.chunks
-                  .map(_.toByteBuffer)
-                  .toPublisherResource
-                  .use(pub => s3.upload(uploadBucket, id.toString, FlowAdapters.toPublisher(pub)))
-                  .flatMap { res =>
-                    checksumResult.map { checksum =>
+                stream.compile.toList.flatMap { bytes =>
+                  Stream
+                    .emit[IO, ByteBuffer](ByteBuffer.wrap(bytes.toArray))
```
techncl, line 179: Since you use it twice, you could save `bytes.toArray` to a variable

techncl (commented):
Looks good! Just one comment

scala/lambdas/ingest-parsed-court-document-event-handler/src/main/scala/uk/gov/nationalarchives/ingestparsedcourtdocumenteventhandler/FileProcessor.scala
```diff
@@ -174,22 +173,17 @@ class FileProcessor(
...
+                    .toPublisherResource
+                    .use(pub => s3.upload(uploadBucket, id.toString, FlowAdapters.toPublisher(pub)))
+                    .map { _ =>
+                      val checksum = DigestUtils.sha256Hex(bytes.toArray)
                       tarEntry.getName -> FileInfo(id, tarEntry.getSize, tarEntry.getName.split('/').last, checksum, URI.create(s"s3://$uploadBucket/${id.toString}"))
```
sparkhi, line 184: As we have now started supporting multiple checksums elsewhere, it might be useful to modify the `FileInfo` case class within this to call the checksum as `sha256Checksum` so we know, in this type of ingest we are generating only one checksum and it is always `SHA256` 

`case class FileInfo(id: UUID, fileSize: Long, fileName: String, sha256Checksum: String, location: URI)`

MancunianSam: Good point. I've changed it in the case class and changed the variable name on this line.

techncl (approved):
👍

PR #247, opened 2024-09-17
Encode MessageType correctly.

MancunianSam:
This is being printed as
```json
"type": {
 "IngestUpdate": {}
 }
```
This is unhelpful, it should be
```json
"type": "preserve.digital.asset.ingest.update"
```

This should fix it


techncl (approved):
👍

PR #249, opened 2024-09-18
DR2-1884 Change handler not sending update message

MancunianSam:
The problem is with the logic in the change handler and the order we get sent the messages.

Because the custodial-copy process picks up the ingested files before the step function sets ingested_PS to true, we get two updates in this order

```
ingested_PS=false, ingested_CC=true
ingested_PS=true,ingested_CC= true`
```

The logic was:

```scala
if asset.ingestedPreservica && asset.ingestedCustodialCopy then
  //send IngestComplete
else if asset.ingestedPreservica && asset.skipIngest then
  //Send IngestUpdate with some other logic around assets with the same Id
else if asset.ingestedPreservica then
 // Send ingest update message
So it was never hitting the last else if because ingestedPreservica was never true on its own.
```

The solution is to split them:

```scala
if asset.ingestedPreservica then
  //send IngestComplete
else if asset.ingestedPreservica && asset.skipIngest then
  //Send IngestUpdate with some other logic around assets with the same Id

if asset.ingestedPreservica && asset.ingestedCustodialCopy then
 // Send ingest update message
```

This means that we’ll get an IngestUpdate message and an IngestComplete message.

If we do sometimes get

```
ingested_PS=true, ingested_CC=false
ingested_PS=true,ingested_CC= true
```

Then we’ll get two IngestUpdate messages and one IngestComplete message. We can live with this, both because it’s unlikely and because anything downstream should be idempotent. It’s the IngestComplete message that other teams will react to and we’ll only get one of those.

### DIFF
--- a/scala/lambdas/ingest-files-change-handler/TestCases.md
+++ b/scala/lambdas/ingest-files-change-handler/TestCases.md
@@ -401,6 +401,20 @@ Custodial Copy, where all the child Files are downloaded to Custodial Copy.
 }
 ```
 
+```json
+{
+  "properties": {
+    "messageId": "e2f8a9a9-4935-4f4f-96f8-7bdfa75960e3",
+    "parentMessageId": null,
+    "timestamp": "2024-05-23T10:13:16.925Z",
+    "type": "preserve.digital.asset.ingest.update"
+  },
+  "parameters": {
+    "assetId": "1"
+  }
+}
+```
+
 ## 6. `ingested_PS`+`ingested_CC` for 1 Asset only where all files are not complete
 
 This test models the action of adding `ingested_PS` or `ingested_CC` to the Asset after reconciling/downloading to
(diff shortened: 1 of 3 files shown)


techncl:
> The solution is to split them:
> 
> ```scala
> if asset.ingestedPreservica then
>   //send IngestComplete
> else if asset.ingestedPreservica && asset.skipIngest then
>   //Send IngestUpdate with some other logic around assets with the same Id
> 
> if asset.ingestedPreservica && asset.ingestedCustodialCopy then
>  // Send ingest update message
> ```

What the `//` comments state, above, is different to what is in the code; the code makes sense so I'm guessing what you added here was just a copy and paste error.

> This means that we’ll get an IngestUpdate message and an IngestComplete message.

Just out of curiosity, could this be avoided by adding `!asset.ingestedCustodialCopy`, like 

```
if asset.ingestedPreservica && !asset.skipIngest && !asset.ingestedCustodialCopy then sendOutputMessage(asset, IngestUpdate)

else if asset.ingestedPreservica && asset.skipIngest && !asset.ingestedCustodialCopy then
```?

techncl (approved):
Looks good!

PR #250, opened 2024-09-18
DR2-1876 missing parent errors.

MancunianSam:
We have been getting an error:

`Expected 1 parent asset, found 0 assets for file <id>: java.lang.Exception`

This is most likely because the update to the asset has been processed first and the asset has been deleted.

The solution is just to ignore it if the file has no parent.

This is because if the file has no parent, this means that the parent has already been processed and the ingest complete message sent.

We can’t ever get into a race condition if the asset is complete and the files aren’t because the ingest complete message won’t send for the asset if that’s true and the asset will still be there.

I’ve changed the method which gets the parent asset to return an `Option`. If there isn’t one, it will do nothing.

I've removed the test which checks for this error and I've added one in for a missing parent path which should never happen.

### DIFF
--- a/scala/lambdas/ingest-files-change-handler/src/main/scala/uk/gov/nationalarchives/ingestfileschangehandler/Lambda.scala
+++ b/scala/lambdas/ingest-files-change-handler/src/main/scala/uk/gov/nationalarchives/ingestfileschangehandler/Lambda.scala
@@ -57,7 +57,7 @@ class Lambda extends LambdaRunner[DynamodbEvent, Unit, Config, Dependencies]:
         .queryItems[FileDynamoTable](config.dynamoTableName, "batchId" === asset.batchId and "parentPath" === childrenParentPath, Option(config.dynamoGsiName))
     }
 
-    def getParentAsset(fileRow: FileDynamoTable): IO[AssetDynamoTable] = for {
+    def getParentAsset(fileRow: FileDynamoTable): IO[Option[AssetDynamoTable]] = for {
       parentId <- IO.fromOption {
         fileRow.potentialParentPath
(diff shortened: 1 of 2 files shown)


techncl:
Looks good, just two questions
> The solution is just to ignore it if the file has no parent.
> 
> This is because if the file has no parent, this means that the parent has already been processed and the ingest complete message sent.

We added this exception initially because we assumed that it was a possibility, so could getting rid of it potentially hide an error (e.g. something like the parent asset being accidentally deleted/deleted too early due to a mistake of ours)? What if we added an attribute to the File that was `AssetIngested` == true, check that and throw an error if it doesn't exist **and** the asset is not present in the table?

> We can’t ever get into a race condition if the asset is complete and the files aren’t because the ingest complete message won’t send for the asset if that’s true and the asset will still be there.

Just to confirm, you're saying here that an Asset can't be marked as "completed" (PS + CC) and deleted if the files haven't been marked as completed and therefore an Asset should be in the table. Does this mean that this error that occurred was due to the fact that the Asset **and** its files were to be deleted, the Asset got deleted first and then by the time it got to the files, the Asset was gone?

MancunianSam:
> What if we added an attribute to the File that was AssetIngested == true, check that and throw an error if it doesn't exist

This makes everything  more complicated for an error that should never exist with proper testing. It's a bit premature optimisation. We know that having it there causes the error and that removing it should be fairly safe unless we make a really bad error so it's worth removing.

> . Does this mean that this error that occurred was due to the fact that the Asset and its files were to be deleted, the Asset got deleted first and then by the time it got to the files, the Asset was gone?

Yes

techncl (approved):
👍

PR #259, opened 2024-09-27
Refactor dynamo formatters

techncl:


MancunianSam (approved):
I've not checked every single line as there's a lot but it looks like a fairly straight rename and the new Dynamo test error check looks sensible.

PR #262, opened 2024-09-30
Write folders and assets UUIDs to JSON and upload to S3

techncl:


MancunianSam (commented):
The main code looks good. 

For the tests, you've now got a weird hybrid of Wiremock and properly created interfaces. 

You can restructure them so that the `Ref` passed to the mock S3 method contains case classes of the objects in the bucket, you can then add/read from this list as the upload and download methods are called. 

You can do a similar thing with the DADynamoDb class which means you can get rid of Wiremock completely. 

I've put a branch in to show you what I mean. What do you think? 

https://github.com/nationalarchives/dr2-ingest/compare/DR2-1781_writeItemUuidsToS3...write-item-uuids-suggested-changes?expand=1

techncl:
> https://github.com/nationalarchives/dr2-ingest/compare/DR2-1781_writeItemUuidsToS3...write-item-uuids-suggested-changes?expand=1

Sure, looks good! I was originally planning to do remove Wiremock from this file on this PR but thought that I should maybe do it on a separate one but I'll do it here.

MancunianSam (approved):
Cool, thanks. I wouldn't normally have written it to get rid of wiremock but I was trying to get out of doing the documentation.

PR #267, opened 2024-10-02
Dr2 1858 validate incoming tdr metadata

sparkhi:
Changes:
1) Copy the object and metadata AFTER the validation has succeeded 
2) Validation -  Mandatory fields (TransferIniitiatedDate, Series, UUID)
3) Validation - format of UUID, Date and Making sure series is not empty
4) Associated unit tests


python/lambdas/copy-files-from-tdr/lambda_function.py
```diff
@@ -1,23 +1,79 @@
...
         source_bucket = body['bucket']
+        assert_objects_exist(source_bucket, file_id)
+        validate_metadata(source_bucket, f"{file_id}.metadata")
         file_location = copy_object_to_s3(destination_bucket, file_id, source_bucket)
         copy_object_to_s3(destination_bucket, f"{file_id}.metadata", source_bucket)
         sqs_body = json.dumps({'id': file_id, 'location': file_location})
         sqs_client.send_message(QueueUrl=destination_queue, MessageBody=sqs_body)
 
+def assert_objects_exist(source_bucket, file_id):
+    try:
```
techncl, line 31: Instead of repeating the trys, could we just add the two files to a List, pass them in and then for loop on them?

python/lambdas/copy-files-from-tdr/lambda_function.py
```diff
@@ -1,23 +1,79 @@
...
+
+
+def validate_formats(bucket, s3_key):
+    response = s3_client.get_object(Bucket=bucket, Key=s3_key)
+    json_content = json.loads(response['Body'])
+    uuid_content = json_content['UUID']
+    try:
+        uuid_object = uuid.UUID(uuid_content)
+    except ValueError:
+        raise Exception(f"Unable to parse UUID, '{uuid_content}'. Invalid format")
```
techncl, line 49: Should we add the file name to this error message?

python/lambdas/copy-files-from-tdr/lambda_function.py
```diff
@@ -1,23 +1,79 @@
...
+
+    transfer_initiated_date_content = json_content['TransferInitiatedDatetime']
+    try:
+        transfer_initiated_date =  datetime.strptime(transfer_initiated_date_content, "%Y-%m-%d %H:%M:%S")
+    except ValueError:
+        raise Exception(f"Unable to parse date, '{transfer_initiated_date_content}'. Invalid format")
```
techncl, line 55: Should we add the file name to this error message?

python/lambdas/copy-files-from-tdr/copy_files_lambda_function_test.py
```diff
@@ -138,5 +156,135 @@ def test_send_message_failure(self, mock_send_message, _, mock_head_object):
...
+        mock_head_object.side_effect = client_error
+        with self.assertRaises(Exception) as ex:
+            lambda_function.assert_objects_exist('some_bucket', 'some_key')
+
+        self.assertEqual('Object some_key does not exist, underlying error is: '
+                         'An error occurred (404) when calling the HeadObject operation: Not Found' , str(ex.exception))
+
+
+    @patch('lambda_function.s3_client.head_object')
+    def test_should_raise_an_exception_when_the_object_exists_but_metadata_does_not_exist_in_source_bucket(self, mock_head_object):
```
techncl, line 234: I'm not sure that this test name is 100% correct because it says "**when the object exists** but metadata does not exist" but this test does not do anything to verify that the object exists, we only assume that it does because it has reached this stage, but if you were to remove the headObject check in the `assert_objects_exist` method or change the order they were called, this test name wouldn't be true. I think it should just be `test_should_raise_an_exception_when_the_metadata_does_not_exist_in_source_bucket`

MancunianSam (commented):
Looks good. Just a few minor bits. 

We also need to do something with the deployment because this is the first python lambda with a dependency. I think the easiest way is to have a deploy script in the project directory and call that from the build script.

It's a bit fiddly to write in a comment so I've put a branch in so you can see what I mean
https://github.com/nationalarchives/dr2-ingest/compare/DR2-1858-Validate-incoming-TDR-metadata...deploy-changes-for-metadata-validation?expand=1

python/lambdas/copy-files-from-tdr/lambda_function.py
```diff
@@ -1,23 +1,79 @@
...
         file_id = body['fileId']
         source_bucket = body['bucket']
+        assert_objects_exist(source_bucket, file_id)
+        validate_metadata(source_bucket, f"{file_id}.metadata")
         file_location = copy_object_to_s3(destination_bucket, file_id, source_bucket)
         copy_object_to_s3(destination_bucket, f"{file_id}.metadata", source_bucket)
         sqs_body = json.dumps({'id': file_id, 'location': file_location})
         sqs_client.send_message(QueueUrl=destination_queue, MessageBody=sqs_body)
 
+def assert_objects_exist(source_bucket, file_id):
```
techncl, line 30: Maybe `assert_objects_exist_in_bucket`?

techncl (commented):
Looks good, just some minor comments, as well as these:

1. There is a mix of single quotes and double, maybe we could just use double
2. We had a conversation about the schema not being confined to TDR, do you think it's worth not calling it?
3. for the `validate_metadata` errors, do you think it's worth collecting all of them and then throwing an error at the end with all of the messages, just in case a metadata file has more than one error in it?
4. Should we be checking the format of the Series?
5. In `validate_formats`, you do `not series.strip()`, could you add a another test where the Series is a space (or multiple spaces)? That way, if `not series.strip()` is removed/modified, it will throw an error
6. For the first test `test_copy`, could you verify that `assert_objects_exist` and `validate_metadata` are being called, using `self.assertEqual` (there's an example in that test) just in the event of accidental removal/modification of the two lines you added in the `lambda_handler`

python/lambdas/copy-files-from-tdr/lambda_function.py
```diff
@@ -1,23 +1,79 @@
...
-
 def lambda_handler(event, context):
     destination_bucket = os.environ["DESTINATION_BUCKET"]
     destination_queue = os.environ["DESTINATION_QUEUE"]
     for record in event['Records']:
         body = json.loads(record['body'])
         file_id = body['fileId']
         source_bucket = body['bucket']
+        assert_objects_exist(source_bucket, file_id)
+        validate_metadata(source_bucket, f"{file_id}.metadata")
```
kurtismash, line 24: @MancunianSam I don't like that we're inferring a metadata file exists from the message. Can we add tech debt to us and TDR to fix this by including both object refs in the message.

kurtismash: \+ if we are inferring, let's do it once and use it as a variable.

python/lambdas/copy-files-from-tdr/lambda_function.py
kurtismash: I'm not sure the trys add anything here. If the object doesn't exist it'll log `f"Object {file_id} does not exist, underlying error is: {ex}"`, but if any other error happens it'll still log as an unhandled exception.

python/lambdas/copy-files-from-tdr/lambda_function.py
```diff
@@ -1,23 +1,79 @@
...
+    except ValueError:
+        raise Exception(f"Unable to parse date, '{transfer_initiated_date_content}'. Invalid format")
+
+    series = json_content['Series']
+    if not series.strip():
+        raise Exception(f"Empty series value, unable to proceed")
+
+    return True
+
+def validate_metadata(bucket, s3_key):
```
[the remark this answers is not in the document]
kurtismash: \+1

python/lambdas/copy-files-from-tdr/lambda_function.py
```diff
@@ -1,23 +1,79 @@
...
+def validate_formats(bucket, s3_key):
+    response = s3_client.get_object(Bucket=bucket, Key=s3_key)
+    json_content = json.loads(response['Body'])
+    uuid_content = json_content['UUID']
+    try:
+        uuid_object = uuid.UUID(uuid_content)
+    except ValueError:
+        raise Exception(f"Unable to parse UUID, '{uuid_content}'. Invalid format")
+
+    transfer_initiated_date_content = json_content['TransferInitiatedDatetime']
```
kurtismash, line 51: Should this be `Consignment-Completed-Datetime` (field name taken from `bag-info.txt` in the BagIt export?

python/lambdas/copy-files-from-tdr/lambda_function.py
```diff
@@ -1,23 +1,79 @@
...
+    try:
+        transfer_initiated_date =  datetime.strptime(transfer_initiated_date_content, "%Y-%m-%d %H:%M:%S")
+    except ValueError:
+        raise Exception(f"Unable to parse date, '{transfer_initiated_date_content}'. Invalid format")
+
+    series = json_content['Series']
```
kurtismash, line 57: I think we need to do more here to check the series is valid. We infer the department from the series and I'm not sure what the Mapper Lambda will do if that doesn't work.

python/lambdas/copy-files-from-tdr/lambda_function.py
```diff
@@ -1,23 +1,79 @@
...
+    json_content = response['Body']
+    data = json.loads(json_content)
+    try:
+        validate(data, incoming_schema)
+    except jsonschema.exceptions.ValidationError as err:
+        raise Exception(err.message)
+
+    return True
 
 def copy_object_to_s3(destination_bucket, s3_key, source_bucket):
```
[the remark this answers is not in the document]
kurtismash: `copy_object` as the buckets _could_ be the same.

python/lambdas/copy-files-from-tdr/tdr_json_schema.py
kurtismash, line 1: Is there benefit to this being a Python file over a JSON file?

python/lambdas/copy-files-from-tdr/lambda_function.py
techncl: The validator [that I did ](https://github.com/nationalarchives/dr2-ingest/blob/324050bcd31ddf06d9c57beb2a743db04e0bd97d/scala/lambdas/ingest-validate-generic-ingest-inputs/src/main/resources/metadata-validation-schemas/archive-and-content-folder-validation-schema.json#L9) does use Regex to validate it but maybe we want to add it here too?

python/lambdas/copy-files-from-tdr/lambda_function.py
MancunianSam: Yeah fair point, I'll do it soon-ish. I've reached my TDR fiddling quota for the month.

python/lambdas/copy-files-from-tdr/tdr_json_schema.py
```diff
@@ -0,0 +1,12 @@
+import jsonschema
```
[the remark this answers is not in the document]
sparkhi: I had kept it in a separate file as I felt there will be more of these in future, this one is specific to TDR, Happy to move it into the same file

python/lambdas/copy-files-from-tdr/lambda_function.py
```diff
@@ -1,23 +1,79 @@
...
         file_location = copy_object_to_s3(destination_bucket, file_id, source_bucket)
         copy_object_to_s3(destination_bucket, f"{file_id}.metadata", source_bucket)
         sqs_body = json.dumps({'id': file_id, 'location': file_location})
         sqs_client.send_message(QueueUrl=destination_queue, MessageBody=sqs_body)
 
+def assert_objects_exist(source_bucket, file_id):
+    try:
+        s3_client.head_object(source_bucket, file_id)
+    except botocore.exceptions.ClientError as ex:
+        raise Exception(f"Object {file_id} does not exist, underlying error is: {ex}")
```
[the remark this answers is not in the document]
sparkhi: Is that okay? I was very cautious in not including many details in exceptions as it has a potential to reveal internal infrastructure details

python/lambdas/copy-files-from-tdr/lambda_function.py
sparkhi: Typically we only catch those specific exceptions that we can handle, hence a specific catch of ClientError and modifying the exception to include the missing id. If there are other errors thrown (e.g. Endpoint connection error or no credentials error), they are unrelated to existence of the objects, hence they should be bubbled up as they are, isn't it?

python/lambdas/copy-files-from-tdr/lambda_function.py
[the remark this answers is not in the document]
kurtismash: I think so, the logs aren't publicly available so users would need to be authenticated and authorised to view them within our AWS environment. Hiding bucket names is implementing security through obscurity, where the bucket should be protected other means - if a bucket name were published we wouldn't replace the bucket. On top of this, the bucket name is already available through our practice of open code.

python/lambdas/copy-files-from-tdr/lambda_function.py
kurtismash: Yea, that makes sense. I had forgotten that the stock error from AWS doesn't include the bucket and object when I wrote the comment.

python/lambdas/copy-files-from-tdr/copy_files_lambda_function_test.py
sparkhi: I disagree, the mock that is setup below, is setup only to throw an exception on metadata, hence it is precisely doing what the name suggests.

techncl: Yes, sorry if I wasn't clear earlier but what I meant is you have "when_the_object_exists" in the test name but you are not testing that the object exists in this test, you're just testing that the metadata exists

techncl: Maybe I should have asked "what do you mean by 'object' in your test title"?

python/lambdas/copy-files-from-tdr/tdr_json_schema.py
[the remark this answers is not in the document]
MancunianSam: This is just a TDR lambda though, we'll have other ones for other transfer types I think so put it in the same file.

python/lambdas/copy-files-from-tdr/lambda_function.py
sparkhi: Do you mean change the name in TDR code?

kurtismash: What is `TransferInitiatedDatetime` in TDR? cc @MancunianSam

python/lambdas/copy-files-from-tdr/lambda_function.py
sparkhi: Hmmmm I misunderstood. I thought we specifically wanted to avoid restrictive checks, I definitely recall conversations to that effect. "MOCK1 123" is invalid by current convention used in DRI. I can use the same regex as the one from Nigel's code

python/lambdas/copy-files-from-tdr/lambda_function.py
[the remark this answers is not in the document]
sparkhi: Done

python/lambdas/copy-files-from-tdr/copy_files_lambda_function_test.py
sparkhi: Calling it a file now.

python/lambdas/copy-files-from-tdr/lambda_function.py
sparkhi: I'll remove format checking, much easier to remove code than add

python/lambdas/copy-files-from-tdr/copy_files_lambda_function_test.py
```diff
@@ -246,44 +214,46 @@ def selective_error_object(bucket, key) :
...
+        }
         with self.assertRaises(Exception) as ex:
-            lambda_function.validate_formats('any_bucket_patched', 'any_key_patched')
-        self.assertEqual("Empty series value, unable to proceed" , str(ex.exception))
+            lambda_function.validate_formats(mock_get_object_response, 'any_bucket_patched', 'any_key_patched')
+        self.assertEqual(
+            "Empty Series value in file 'any_key_patched' in bucket 'any_bucket_patched'. Unable to proceed",
+            str(ex.exception))
+
+    # def test_should_raise_an_exception_when_series_format_is_invalid(self):
```
techncl, line 246: Did you mean to comment this out?

python/lambdas/copy-files-from-tdr/lambda_function.py
```diff
@@ -1,85 +1,97 @@
...
+    except ValueError:
+        raise Exception(
+            f"Unable to parse UUID, '{uuid_content}' from file '{s3_key}' in bucket '{bucket}'. Invalid format")
+
+    series = json_metadata['Series']
+    if not series.strip():
+        raise Exception(f"Empty Series value in file '{s3_key}' in bucket '{bucket}'. Unable to proceed")
+
+    #FIXME, Can we use this regex pattern? The test series MOCK1 123 fails this pattern
+    # pattern = "^([A-Z]{1,4} [1-9][0-9]{0,3})"
```
techncl, line 71: 1. I know this is a work in progress but it needs a `$` at the end. 
2. also, I [upgraded the regex](https://github.com/nationalarchives/dr2-ingest/blob/main/scala/lambdas/ingest-validate-generic-ingest-inputs/src/main/resources/metadata-validation-schemas/archive-and-content-folder-validation-schema.json) in the validator that I did to include `MOCK1 123`

python/lambdas/copy-files-from-tdr/copy_files_lambda_function_test.py
sparkhi: yes, it is related to FIXME in the implementation

sparkhi:
> Looks good. Just a few minor bits.
> 
> We also need to do something with the deployment because this is the first python lambda with a dependency. I think the easiest way is to have a deploy script in the project directory and call that from the build script.
> 
> It's a bit fiddly to write in a comment so I've put a branch in so you can see what I mean https://github.com/nationalarchives/dr2-ingest/compare/DR2-1858-Validate-incoming-TDR-metadata...deploy-changes-for-metadata-validation?expand=1

Included these changes now

python/lambdas/copy-files-from-tdr/lambda_function.py
sparkhi: Done

MancunianSam (approved):

PR #276, opened 2024-10-07
DR2-1909 Ingest failure notifications

MancunianSam:
This lambda will be triggered by the step function failure
notifications.

It first gets the groupId and batchId out of the step function input
which is sent as part of the eventbridge rule. This will work for the
preingest and ingest step functions as they both have these in their
input.

It then looks up that group id in the lock table and for each asset it
finds, it will send an update message saying there has been an error.

I also renamed EventDecoders to EventCodecs as it has encoders now as
well and I moved some of the messaging stuff into utils as there's
multiple things using it now.

### DIFF
--- a/scala/lambdas/ingest-failure-notifications/src/main/scala/uk/gov/nationalarchives/ingestfailurenotifications/Lambda.scala
+++ b/scala/lambdas/ingest-failure-notifications/src/main/scala/uk/gov/nationalarchives/ingestfailurenotifications/Lambda.scala
@@ -1,0 +1,64 @@
+package uk.gov.nationalarchives.ingestfailurenotifications
+
+import cats.effect.IO
+import io.circe.DecodingFailure.Reason.CustomReason
+import io.circe.generic.semiauto.deriveDecoder
+import io.circe.{Decoder, DecodingFailure, HCursor}
+import io.circe.parser.decode
+import pureconfig.ConfigReader
+import pureconfig.generic.derivation.default.*
+import uk.gov.nationalarchives.{DADynamoDBClient, DASNSClient}
+import uk.gov.nationalarchives.utils.LambdaRunner
+import uk.gov.nationalarchives.ingestfailurenotifications.Lambda.{*, given}
+import uk.gov.nationalarchives.DADynamoDBClient.given
+import org.scanamo.syntax.*
+import uk.gov.nationalarchives.dynamoformatters.DynamoFormatters.IngestLockTableItem
+import uk.gov.nationalarchives.utils.ExternalUtils.MessageStatus.IngestError
+import uk.gov.nationalarchives.utils.EventCodecs.given
+import uk.gov.nationalarchives.utils.ExternalUtils.MessageType.IngestUpdate
+import uk.gov.nationalarchives.utils.ExternalUtils.{OutputMessage, OutputParameters, OutputProperties}
+
+import java.time.Instant
+import java.util.UUID
+
+class Lambda extends LambdaRunner[SfnEvent, Unit, Config, Dependencies]:
+  override def handler: (SfnEvent, Config, Dependencies) => IO[Unit] = (event, config, dependencies) =>
+    for {
+      items <- dependencies.dynamoClient
+        .queryItems[IngestLockTableItem](config.lockTableName, "groupId" === event.detail.input.groupId, Option(config.lockTableGsiName))
+      messages = items.map { item =>
+        OutputMessage(
+          OutputProperties(event.detail.input.batchId, dependencies.uuidGenerator(), None, Instant.now, IngestUpdate),
+          OutputParameters(item.assetId, IngestError)
+        )
+      }
+      _ <- dependencies.snsClient.publish(config.topicArn)(messages)
+    } yield ()
+
+  override def dependencies(config: Config): IO[Dependencies] =
+    IO.pure(Dependencies(DASNSClient[IO](), DADynamoDBClient[IO](), () => UUID.randomUUID))
+end Lambda
+
+object Lambda:
+  given Decoder[SfnInput] = deriveDecoder[SfnInput]
+
+  given Decoder[SfnDetail] = (c: HCursor) =>
+    for {
+      inputString <- c.downField("input").as[String]
+      input <- decode[SfnInput](inputString).left.map(err => DecodingFailure(CustomReason(err.getMessage), c))
+    } yield SfnDetail(input)
+
+  given Decoder[SfnEvent] = (c: HCursor) =>
+    for {
+      detail <- c.downField("detail").as[SfnDetail]
+    } yield SfnEvent(detail)
+
+  case class SfnInput(groupId: String, batchId: String)
+
+  case class SfnDetail(input: SfnInput)
+
+  case class SfnEvent(detail: SfnDetail)
+
+  case class Config(topicArn: String, lockTableName: String, lockTableGsiName: String) derives ConfigReader
+
+  case class Dependencies(snsClient: DASNSClient[IO], dynamoClient: DADynamoDBClient[IO], uuidGenerator: () => UUID)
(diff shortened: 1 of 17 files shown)


scala/lambdas/ingest-failure-notifications/src/test/scala/uk/gov/nationalarchives/ingestfailurenotifications/LambdaTest.scala
```diff
@@ -0,0 +1,65 @@
...
+
+        properties.messageType should equal(MessageType.IngestUpdate)
+        properties.messageId should equal(uuids.head)
+        properties.executionId should equal("groupId_0")
+        properties.parentMessageId should equal(None)
+        checkMessages(snsMessages.tail)
+    }
+  }
+
+  "handler" should "send no messages if there are items in the lock table with a different group id" in {
```
techncl, line 40: Should this not be an error/warning message? Why would an event be triggered if it wasn't in the table? Or could the items be deleted in between the event trigger and the get from dynamo?

scala/lambdas/ingest-failure-notifications/src/test/scala/uk/gov/nationalarchives/ingestfailurenotifications/LambdaTest.scala
```diff
@@ -0,0 +1,65 @@
...
+    val items = generateItems()
+    val snsMessages = runLambda(items, SfnInput("differentGroupId", "groupId_0"))
+
+    snsMessages.size should equal(0)
+  }
+
+  "handler" should "send no messages if there are no items in the lock table" in {
```
techncl, line 47: Same question as above

scala/lambdas/ingest-failure-notifications/src/main/scala/uk/gov/nationalarchives/ingestfailurenotifications/Lambda.scala
```diff
@@ -0,0 +1,64 @@
...
+import uk.gov.nationalarchives.utils.ExternalUtils.{OutputMessage, OutputParameters, OutputProperties}
+
+import java.time.Instant
+import java.util.UUID
+
+class Lambda extends LambdaRunner[SfnEvent, Unit, Config, Dependencies]:
+  override def handler: (SfnEvent, Config, Dependencies) => IO[Unit] = (event, config, dependencies) =>
+    for {
+      items <- dependencies.dynamoClient
+        .queryItems[IngestLockTableItem](config.lockTableName, "groupId" === event.detail.input.groupId, Option(config.lockTableGsiName))
```
techncl, line 28: Should we use the `groupId` variable from Dynamoformatters instead of a String?

scala/lambdas/ingest-failure-notifications/src/main/scala/uk/gov/nationalarchives/ingestfailurenotifications/Lambda.scala
```diff
@@ -0,0 +1,64 @@
...
+      messages = items.map { item =>
+        OutputMessage(
+          OutputProperties(event.detail.input.batchId, dependencies.uuidGenerator(), None, Instant.now, IngestUpdate),
```
techncl, line 31: So does this mean that all items we get from Dynamo (via the `groupId`) will have the same `batchId`?

scala/lambdas/ingest-failure-notifications/src/test/scala/uk/gov/nationalarchives/ingestfailurenotifications/LambdaTestUtils.scala
```diff
@@ -0,0 +1,81 @@
...
+        if dynamoError then IO.raiseError(new Exception("Error getting dynamo items"))
+        else
+          dynamoItemsRef.get.map { rows =>
+            (for {
+              values <- requestCondition.dynamoValues
+              map <- values.toMap[String].toOption
+            } yield rows
+              .filter(row => map.get("conditionAttributeValue0").contains(row.groupId))
+              .map(_.asInstanceOf[U])).getOrElse(Nil)
+
```
techncl, line 46: There's an extra new line here

techncl (commented):
Looks good! Just some minor comments, also, could you add a README?

scala/lambdas/ingest-failure-notifications/src/test/scala/uk/gov/nationalarchives/ingestfailurenotifications/LambdaTest.scala
```diff
@@ -0,0 +1,65 @@
...
+  "handler" should "send one message for each item in the lock table" in {
+    val items = generateItems()
+    val snsMessages = runLambda(items, SfnInput("groupId", "groupId_0"))
+
+    snsMessages.size should equal(items.size)
+
+    checkMessages(snsMessages)
+
+    @tailrec
+    def checkMessages(snsMessages: List[OutputMessage]): Unit = {
```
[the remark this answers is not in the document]
MancunianSam: I think I did it because I needed the same index from the `items` list but it's easier to do `zipWithIndex` and `foreach` so I've changed it to that.

scala/lambdas/ingest-failure-notifications/src/test/scala/uk/gov/nationalarchives/ingestfailurenotifications/LambdaTestUtils.scala
```diff
@@ -0,0 +1,81 @@
...
+            .map(_ => List(PublishBatchResponse.builder.build))
+
+    val uuidIterator: () => UUID = () => {
+      val uuidsIterator: Iterator[UUID] = uuids.iterator
+      uuidsIterator.next()
+    }
+    Dependencies(snsClient, dynamoClient, uuidIterator)
+  }
+
+  def generateItems(): List[IngestLockTableItem] = List.fill(100)(UUID.randomUUID).zipWithIndex.map { (id, idx) =>
```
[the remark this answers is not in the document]
MancunianSam: Good spot, I think it was at some point. I've changed it to `map`

scala/lambdas/ingest-failure-notifications/src/test/scala/uk/gov/nationalarchives/ingestfailurenotifications/LambdaTestUtils.scala
```diff
@@ -0,0 +1,81 @@
...
+    IngestLockTableItem(id, s"groupId", "")
```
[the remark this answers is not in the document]
MancunianSam: The idx I think :smile: I've removed it.

scala/lambdas/ingest-failure-notifications/src/test/scala/uk/gov/nationalarchives/ingestfailurenotifications/LambdaTest.scala
MancunianSam: Same answer as above :smile:

scala/lambdas/ingest-failure-notifications/src/main/scala/uk/gov/nationalarchives/ingestfailurenotifications/Lambda.scala
MancunianSam: What we're sending here is the batchId in the TDR_<uuid>_0 format. Because this is triggered by the step function failure, we know that all the messages sent will be for the batch that step function is processing if that makes sense. 
The batchId is only linked to the lock table in the sense that that's the ID of the currently running step function.

techncl: Ah, so even though we are getting the items from dynamo based on the `groupId`, they will all have the same `batchId` because the locktable on holds what's actually running and therefore you can't items belonging to both `_0` and `_1` batches at the same time?

MancunianSam: Yeah, exactly.

techncl (approved):
👍

PR #279, opened 2024-10-08
Minor copy lambda fixes.

MancunianSam:
We were getting an error
` No module named 'rpds.rpds`

This issue comment
https://github.com/aws/aws-cdk/issues/26300#issuecomment-1629842430

says we need to pin the jsonschema import to that version. This has
stopped the error.

Also, head_object in boto3 needs named arguments and I've updated the
tests to reflect this.


techncl (approved):
👍

PR #280, opened 2024-10-09
DR2-1919 Add correlationId to schema & refactor 'optional properties'  tests

techncl:
This property is optional


MancunianSam (approved):
:+1:

PR #281, opened 2024-10-09
DR2 1907 remove content folders assets from output

MancunianSam:
- **DR2-1907 Remove content assets and content folders from output.**
- **Remove comment**


techncl (approved):
👍

PR #284, opened 2024-10-10
Dr2 1906 remove asset checks from workflow monitor

MancunianSam:


MancunianSam:
The README is accurate now I think. It was never updated to say that we were getting the assets from the messages.

techncl (approved):
👍

PR #286, opened 2024-10-11
Dr2 1786 remove name from assets

sparkhi:
Main change, removed 'name' from `AssetDynamoItem` 
As it was interlinked with other classes, class hierarchy changed between `asset` and other items.
Fixed compilation errors and test failures


techncl (commented):
Looks good! Would you mind also editing the README for the Reconciler and "Find existing asset" as those should need updating?

techncl (approved):
👍

PR #287, opened 2024-10-14
Dr2 1911 dr2 1912 add correlation id to asset in metadata package

techncl:


MancunianSam (approved):
I've fixed your conflct as it was my changes they were clashing with. I was also getting weird compile errors from the LockTableMessage type alias in the test class so I just replaced it with the original case class.

PR #289, opened 2024-10-14
DR2-1870 E2E tests

MancunianSam:
There are two types of test.

Full end to end : This sends a message to the copy-files-from-tdr queue which triggers an
ingest just as the TDR export would. This is used for the happy path
test and the test that fails on the TDR validation.

Manually adding to the lock table: For tests which will fail, we need to control which batch they appear in
and the only way to do this is to fake the job of the aggregator and
start the preingest step function manually.

For the test that fails the TDR validation, we check the
copy-files-from-tdr-dlq. For all others, there is an e2e-tests queue
subscribed to the notiffications topic on integration only and we poll
this for the success and failure messages.

I've not implemented judgments tests because I would have to generate a
tar.gz package and I don't want to.

We also wanted BDD style feature files but I didn't want Cucumber so
I've implemented our own version here which is a bit more cats effect
friendly.


scala/e2e-tests/src/test/scala/uk/gov/nationalarchives/e2etests/StepDefs.scala
```diff
@@ -0,0 +1,193 @@
...
+  def createMetadataJson(id: UUID, checksum: String, invalidMetadata: Boolean): Array[Byte] =
+    def generateValue[T](ifPresent: T) = if invalidMetadata && Random.nextInt(100) > 50 then None else ifPresent.some
+    val series = if invalidMetadata then None else Some("TEST 123")
+    TDRMetadata(
+      series,
+      generateValue(id),
+      None,
+      "TestBody",
+      generateValue("2024-10-07 09:54:48"),
+      s"TST-${LocalDate.now.getYear}-${makeReference(4)}",
```
techncl, line 86: 1. Could `LocalDate.now.getYear` be a global private val?
2. I thought we were already using the prefix `TST`; should we use something else like `E2E` (in order to differentiate)? Same question for line 47

scala/e2e-tests/src/test/scala/uk/gov/nationalarchives/e2etests/StepDefs.scala
```diff
@@ -0,0 +1,193 @@
...
+      .map(_.toByteBuffer)
+      .toPublisherResource.use { pub =>
+        s3Client.upload(bucket, key, FlowAdapters.toPublisher(pub))
+      }
+
+  def makeReference(length: Int): String =
+    Random.shuffle(('A' to 'Z') ++ ('0' to '9')).take(length).mkString
+
+  def createMetadataJson(id: UUID, checksum: String, invalidMetadata: Boolean): Array[Byte] =
+    def generateValue[T](ifPresent: T) = if invalidMetadata && Random.nextInt(100) > 50 then None else ifPresent.some
```
techncl, line 78: 1. Maybe `ifValuePresent` instead of `ifPresent`?
2. Even though what you have is fine, you could use `Random.nextBoolean()`.
3. Wait, why do we need `&& Random.nextInt(100) > 50`? Wouldn't that mean that if `invalidMetadata` is true, it will only return invalid metadata 50% of the time?
4. As well as `None,` should we add more values that could fail so `Random.shuffle(List(None, Some("NoSpace"), Some(""))).head`?

scala/e2e-tests/src/test/scala/uk/gov/nationalarchives/e2etests/StepDefs.scala
```diff
@@ -0,0 +1,193 @@
...
+    val series = if invalidMetadata then None else Some("TEST 123")
```
techncl, line 79: As well as None, should we add more values that could fail so 

`Random.shuffle(List(None, Some("NoSpace"), Some(""))).head` ?

scala/e2e-tests/src/test/scala/uk/gov/nationalarchives/e2etests/StepDefs.scala
```diff
@@ -0,0 +1,193 @@
...
+  def pollForMessages[T <: SqsMessage](filter: (SqsMessage, List[UUID]) => Boolean, ref: Ref[F, List[UUID]])(using decoder: Decoder[T], queueUrl: String): F[Unit] = {
+    for {
+      uuids <- ref.get
+      messages <- sqsClient.receiveMessages[T](queueUrl)
+      newMessages <- messageUuidCell.updateAndGet(_ ++ messages.map(_.message))
+      _ <- messages.traverse(msg => sqsClient.deleteMessage(queueUrl, msg.receiptHandle))
+      newInputUuids <- ref.updateAndGet { uuids =>
+        uuids.diff(newMessages.filter(msg => filter(msg, uuids)).map(_.getId))
+      }
+      res <- if newInputUuids.isEmpty then
```
techncl, line 146: `res` doesn't seem to be used

scala/e2e-tests/src/test/scala/uk/gov/nationalarchives/e2etests/StepDefs.scala
```diff
@@ -0,0 +1,193 @@
...
+        Async[F].sleep(60.seconds) >> pollForMessages(filter, ref)
+      else
+        pollForMessages(filter, ref)
+    } yield ()
+  }
+}
+
+object StepDefs {
+  def apply[F[_]: Async](messageUuidCell: AtomicCell[F, List[SqsMessage]]) = new StepDefs(messageUuidCell)
+
```
techncl, line 158: There's an extra new line here

scala/e2e-tests/src/test/scala/uk/gov/nationalarchives/e2etests/StepDefs.scala
```diff
@@ -0,0 +1,193 @@
...
+
+  case class Config(inputBucket: String, ingestSqsQueue: String, notificationsSqsQueue: String, lockTableName: String, preingestSfnArn: String, copyFilesDlq: String)derives ConfigReader
```
techncl, line 160: There's a missing space after the `)`

scala/e2e-tests/src/test/scala/uk/gov/nationalarchives/e2etests/StepDefs.scala
```diff
@@ -0,0 +1,193 @@
...
+import scala.util.Random
+
+class StepDefs[F[_]: Async](messageUuidCell: AtomicCell[F, List[SqsMessage]]) {
+
+  val s3Client: DAS3Client[F] = DAS3Client[F]()
+  val sqsClient: DASQSClient[F] = DASQSClient[F]()
+  val dynamoClient: DADynamoDBClient[F] = DADynamoDBClient[F]()
+  val sfnClient: DASFNClient[F] = DASFNClient[F]()
+  val config: Config = ConfigSource.default.loadOrThrow[Config]
+  val ingestComplete = "preserve.digital.asset.ingest.complete"
```
techncl, line 37: Maybe `ingestCompleteMessage` or `ingestCompleteMsg`?

scala/e2e-tests/src/test/scala/uk/gov/nationalarchives/e2etests/StepDefs.scala
```diff
@@ -0,0 +1,193 @@
...
+  val ingestUpdate = "preserve.digital.asset.ingest.update"
```
techncl, line 38: Maybe `ingestUpdateMessage` or `ingestUpdateMsg`?

techncl (commented):
This looks really good! Just some minor comments as well as:

1. Does/will`sbt scalafmtCheckAll` run for this because I don't think it did this time around?
2. Do you think it's worth adding unit tests for the testRunner and StepDefs as there is no guarantee that this is doing the correct thing?

scala/e2e-tests/src/test/scala/uk/gov/nationalarchives/e2etests/IngestTestsRunner.scala
```diff
@@ -0,0 +1,77 @@
...
+import scala.jdk.CollectionConverters.*
+
+trait IngestTestsRunner extends AnyFunSpec with ParallelTestExecution {
+
+  def runIngestTests(): Unit =
+    Files.list(Path.of(getClass.getResource("/features").getPath))
+      .toList.asScala
+      .map(Files.readString)
+      .toList
+      .flatMap { eachString =>
```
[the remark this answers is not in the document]
MancunianSam: `feature` is good

scala/e2e-tests/src/test/scala/uk/gov/nationalarchives/e2etests/StepDefs.scala
techncl: Could you use `IngestComplete` from the `MessageType` Enum in `ExternalUtils`?

scala/e2e-tests/src/test/resources/features/ingest.feature
```diff
@@ -0,0 +1,23 @@
+Scenario Ingest should succeed if all metadata is valid
```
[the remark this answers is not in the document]
MancunianSam: 1. Yep, it only prints this to console anyway
2. Yeah, I'll trim the lines before parsing them
3. Why not, I can print and ignore it.

scala/e2e-tests/src/test/scala/uk/gov/nationalarchives/e2etests/IngestTestsRunner.scala
```diff
@@ -0,0 +1,77 @@
...
+                    case "When" => testWhen(testString)
+                    case "Then" => testThen(testString)
+                    case invalid => IO.raiseError(new Exception(s"Invalid step $invalid"))
+                case Nil => IO.unit
+            }.void
+          }
+        }
+      }.head
+
+  def ingest(specText: String)(testFun: Ref[IO, List[UUID]] ?=> IO[Unit]): Unit =
```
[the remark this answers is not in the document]
MancunianSam: It's sort of running a test function so scalatest recognises it as a test. I'll call it `runTestFunction` it's not exactly right but it's better than ingest.

scala/e2e-tests/src/test/scala/uk/gov/nationalarchives/e2etests/IngestTestsRunner.scala
```diff
@@ -0,0 +1,77 @@
...
+      .map(Files.readString)
+      .toList
+      .flatMap { eachString =>
+        val eachTest = eachString.split("\n\n")
+        eachTest.toList.map { test =>
+          val eachTestLine = test.split("\n").toList
+          ingest(eachTestLine.head) {
+            eachTestLine.tail.traverse { line =>
+              line.split(" ").toList match
+                case head :: next =>
```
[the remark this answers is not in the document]
MancunianSam: I've used `keyword :: expressionParts` so then `testString` can be changed to `expression`

scala/e2e-tests/src/test/scala/uk/gov/nationalarchives/e2etests/IngestTestsRunner.scala
```diff
@@ -0,0 +1,77 @@
...
+  def runIngestTests(): Unit =
+    Files.list(Path.of(getClass.getResource("/features").getPath))
+      .toList.asScala
+      .map(Files.readString)
+      .toList
+      .flatMap { eachString =>
+        val eachTest = eachString.split("\n\n")
+        eachTest.toList.map { test =>
+          val eachTestLine = test.split("\n").toList
+          ingest(eachTestLine.head) {
```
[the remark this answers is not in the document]
MancunianSam: The `ingest(eachTestLine.head) {` block is difficult to move. I tried and it stopped registering as a test. I can move the block inside it to a method though and I've split some other parts of this out as well.

scala/e2e-tests/src/test/scala/uk/gov/nationalarchives/e2etests/IngestTestsRunner.scala
```diff
@@ -0,0 +1,77 @@
...
+                case Nil => IO.unit
+            }.void
+          }
+        }
+      }.head
+
+  def ingest(specText: String)(testFun: Ref[IO, List[UUID]] ?=> IO[Unit]): Unit =
+    it(specText) {
+      Ref.of[IO, List[UUID]](Nil).flatMap(ref => testFun(using ref)).unsafeRunSync()
+      }
```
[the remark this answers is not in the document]
MancunianSam: I hadn't run scalafmt. I'm going to let it sort this out.

scala/e2e-tests/src/test/scala/uk/gov/nationalarchives/e2etests/IngestTestsRunner.scala
```diff
@@ -0,0 +1,77 @@
...
+            eachTestLine.tail.traverse { line =>
+              line.split(" ").toList match
+                case head :: next =>
+                  val testString = next.mkString(" ")
+                  head match
+                    case "Given" => testGiven(testString)
+                    case "When" => testWhen(testString)
+                    case "Then" => testThen(testString)
+                    case invalid => IO.raiseError(new Exception(s"Invalid step $invalid"))
+                case Nil => IO.unit
```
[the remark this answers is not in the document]
MancunianSam: I will raise an error as a sanity check but I don't think we can ever actually get here now I'm trimming the string because even passing in empty string will stiill yield a list of size one.

scala/e2e-tests/src/test/scala/uk/gov/nationalarchives/e2etests/StepDefs.scala
```diff
@@ -0,0 +1,193 @@
...
+      checksum,
+      s"Z${makeReference(5)}"
+    ).asJson.noSpaces.getBytes
+
+  def createFiles(number: Int, emptyChecksum: Boolean = false, invalidMetadata: Boolean = false, invalidChecksum: Boolean = false)(using ref: Ref[F, List[UUID]]): F[Unit] = {
+    val invalidChecksumValue = "e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855"
+    val ids = List.fill(number)(UUID.randomUUID)
+    for {
+      _ <- Logger[F].info(s"$number ids created")
+      ids <- ids.traverse { id =>
```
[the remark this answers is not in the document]
MancunianSam: `createdFileIds` is OK, I've changed it.

scala/e2e-tests/src/test/scala/uk/gov/nationalarchives/e2etests/StepDefs.scala
MancunianSam: 1. Yeah, good point. I've added it to a variable.
2. That makes sense, I've changed both of them to E2E

scala/e2e-tests/src/test/scala/uk/gov/nationalarchives/e2etests/StepDefs.scala
MancunianSam: 1. There's value in the method name, I don't thinkit needs duplicating here.
2. `nextBoolean` is a good idea, I didn't know that.
3. It always has an empty Series for the invalid metadata because as you say, half the time it would be valid. It just randomly assigns other things to be empty as well. 
4. I've added that in for the series because that's where most of the validation is anyway.

scala/e2e-tests/src/test/scala/uk/gov/nationalarchives/e2etests/StepDefs.scala
MancunianSam: See above, I'll only do it for the series, considering we're not checking why the validation failed I think this is enough.

scala/e2e-tests/src/test/scala/uk/gov/nationalarchives/e2etests/StepDefs.scala
MancunianSam: It's a message type so I'll call it `ingestCompleteType` and `ingestUpdateType` as well.

scala/e2e-tests/src/test/scala/uk/gov/nationalarchives/e2etests/StepDefs.scala
MancunianSam: See above

MancunianSam: I can't use ExternalUtils easily because it's in a seperate subproject with its own build sbt. I quite like that it's seperate though, if we change it the tests will fail which feels right.

techncl (approved):

PR #290, opened 2024-10-15
Add new property instead of determining metadata entries based on name

techncl:


scala/lambdas/ingest-validate-generic-ingest-inputs/src/test/scala/uk/gov/nationalarchives/ingestvalidategenericingestinputs/MetadataJsonValueValidatorTest.scala
```diff
@@ -44,6 +44,9 @@ class MetadataJsonValueValidatorTest extends AnyFlatSpec with MockitoSugar with
     ("Files", Map(3 -> 0, 4 -> 0), "ArchiveFolder")
   )
 
+  private val casesWhereEntryShouldBeLabelledAsMetadataFile: TableFor2[Boolean, String] =
+    Table(("metadataFile value is true", "id is in 'originalMetadataFiles' array"), (true, "is"), (true, "is not"), (false, "is"))
```
MancunianSam, line 48: Can you not use `true` and `false` instead of `is` and `is not` I know it slightly messes up your test name but it makes the code a bit easier to parse?

scala/lambdas/ingest-validate-generic-ingest-inputs/src/test/scala/uk/gov/nationalarchives/ingestvalidategenericingestinputs/MetadataJsonValueValidatorTest.scala
```diff
@@ -44,6 +44,9 @@ class MetadataJsonValueValidatorTest extends AnyFlatSpec with MockitoSugar with
     ("Files", Map(3 -> 0, 4 -> 0), "ArchiveFolder")
   )
 
+  private val casesWhereEntryShouldBeLabelledAsMetadataFile: TableFor2[Boolean, String] =
```
MancunianSam, line 47: This name is far too long. `entriesLabelledAsMetadataFile` perhaps or something similar

scala/lambdas/ingest-validate-generic-ingest-inputs/src/test/scala/uk/gov/nationalarchives/ingestvalidategenericingestinputs/MetadataJsonValueValidatorTest.scala
```diff
@@ -259,38 +262,43 @@ class MetadataJsonValueValidatorTest extends AnyFlatSpec with MockitoSugar with
...
-    "within the Asset's 'originalMetadataFiles' array" in {
-      val entriesWithMetadataFileWithNoExt = allEntries.map { entry =>
-        if entry(id).str == "d4f8613d-2d2a-420d-a729-700c841244f3" then Obj.from(entry.value ++ Map(name -> Str("TDD-2023-ABC-metadata"))) else entry
-      }
-      val entriesGroupedByType = entriesWithMetadataFileWithNoExt.groupBy(_(entryType).str)
-      val entriesAsValidatedMap = entriesGroupedByType.map { case (entryType, entries) =>
-        entryType -> entries.map(convertUjsonObjToSchemaValidatedMap)
-      }
+  forAll(casesWhereEntryShouldBeLabelledAsMetadataFile) { (metadataFileValueIsTrue, isInOriginalMetadataFiles) =>
+    "getIdsOfAllEntries" should s"still assign a file to a 'MetadataFileEntry' type if the 'metadataFile' property is '$metadataFileValueIsTrue' but its id $isInOriginalMetadataFiles contained " +
```
MancunianSam, line 266: The test name is quite hard to parse as it's so long as well so if you can shorten it a bit. If you try and keep it on one line that makes is easier to read.

scala/lambdas/ingest-validate-generic-ingest-inputs/src/test/scala/uk/gov/nationalarchives/ingestvalidategenericingestinputs/MetadataJsonValueValidatorTest.scala
techncl: Hmm, it doesn't quite describe what it is but I could name it `metadataFileCases`?

kurtismash:
Hi @techncl, can you explain a bit more about the benefit this change delivers? What is the purpose of the new field and where are we acting upon it? It feels like an anti-pattern to add constraints if the only outcome is to add a validation check comparing 2 fields in the same input.

techncl:
> Hi @techncl, can you explain a bit more about the benefit this change delivers? What is the purpose of the new field and where are we acting upon it? It feels like an anti-pattern to add constraints if the only outcome is to add a validation check comparing 2 fields in the same input.

Hi @kurtismash, the purpose of this change was mainly to get rid of the reliance on "-metadata.json" to determine whether something is a metadata file; so in my mind, it made sense to label something a metadata file at the point of creation. The reason for labelling something as a metadata file is:

1. titles are allowed to be empty for metadata files but not regular files, so by knowing what is a metadata file, we can validate this
2. We can rely on the Asset's `originalMetadataFiles` array to tell us which file is a metadata file but it's possible that in the future we make a mistake and put a non-metadata file id is entered into that array or we put the metadataId in the `originalFiles` 
3. There is also the possibility that a metadata id is missing from the `originalMetadataFiles` array; there is a test that finds all of the metadata entries and checks if they are present in `originalMetadataFiles`; this can only be done if you can separate the metadataFiles from the non-metadata files
4. Metadata files are checked to see if they end with `.json`

scala/lambdas/ingest-validate-generic-ingest-inputs/src/test/scala/uk/gov/nationalarchives/ingestvalidategenericingestinputs/MetadataJsonValueValidatorTest.scala
MancunianSam: Yeah ok :+1:

kurtismash:
Thanks @techncl. I don't think we should be treating metadata files differently from other files we receive as the role of the service is to preserve the records transferred to us and not to introduce restrictions in that process.

On your point 1, do we need a `title` for any file? It doesn't look like we're using this field within the Asset OPEX Creator, [just `name`](https://github.com/nationalarchives/dr2-ingest/blob/50f5a28b086b7996383f878f14d1500657b79258/scala/lambdas/ingest-asset-opex-creator/src/main/scala/uk/gov/nationalarchives/ingestassetopexcreator/XMLCreator.scala#L139).

I do think there is some elegance in adding `metadataFile` as a property of the File type, but wouldn't want this to duplicate the required input fields; it's like having `title` and `confirmTitle`. If we were to introduce this field I'd like to remove the `originalFiles` and `originalMetadataFiles` fields from the asset type, using the Mapper Lambda to create these when writing to our dr2-ingest-files table. But this gets more difficult as we use these fields only within our custom DR2 Source metadata fragment, which I'd like to keep as a passthrough or configurable mapping within the generic ingest workflow - I think we'd need some XSLT/JSONata/etc. transform within the Mapper to achieve this.

MancunianSam:
So I think the general gist here is that this is a good idea and we'll probably use this instead of the originalFiles and originalMetadataFiles.

There is a bit of work in the opex creators to do before that happens though so it's not a quick job. Also, there's some hefty changes to the validator going in which wil cause problems with this PR. 

We can close this one, and I'll put a ticket in to do this again with the opex creator changes and the updates to the new validator at some point.

PR #297, opened 2024-10-17
Changed the asset subtype to be nullable.

sparkhi:
- Changed from String to option[String]
- corresponding changes to compilations and test successes 
- Added one test for opex xml creator to make sure null is written as empty element


scala/lambdas/dynamo-formatters/src/main/scala/uk/gov/nationalarchives/dynamoformatters/DynamoWriteUtils.scala
```diff
@@ -43,7 +43,6 @@ object DynamoWriteUtils {
           "transferCompleteDatetime" -> DynamoValue.fromString(assetDynamoItem.transferCompleteDatetime.toString),
           "upstreamSystem" -> DynamoValue.fromString(assetDynamoItem.upstreamSystem),
           "digitalAssetSource" -> DynamoValue.fromString(assetDynamoItem.digitalAssetSource),
-          "digitalAssetSubtype" -> DynamoValue.fromString(assetDynamoItem.digitalAssetSubtype),
           "originalFiles" -> DynamoValue.fromStrings(assetDynamoItem.originalFiles.map(_.toString)),
```
MancunianSam, line 46: I think this still needs to be in here. We still want to write it to Dynamo if it's set. 
Can you add it to the bottom of this method like the one for correlationId
```scala
++ assetDynamoItem.potentialDigitalAssetSubtype.map(subType => Map("digitalAssetSubtype" -> DynamoValue.fromString(subType))).getOrElse(Map())
```

scala/lambdas/dynamo-formatters/src/test/scala/uk/gov/nationalarchives/dynamoformatters/DynamoFormattersTest.scala
```diff
@@ -529,7 +530,6 @@ class DynamoFormattersTest extends AnyFlatSpec with TableDrivenPropertyChecks wi
     resultMap(transferCompleteDatetime).s() should equal("2023-06-01T00:00Z")
     resultMap(upstreamSystem).s() should equal(upstreamSystem)
     resultMap(digitalAssetSource).s() should equal(digitalAssetSource)
-    resultMap(digitalAssetSubtype).s() should equal(digitalAssetSubtype)
     resultMap(originalFiles).ss().asScala.toList should equal(List(originalFilesUuid.toString))
     resultMap(originalMetadataFiles).ss().asScala.toList should equal(List(originalMetadataFilesUuid.toString))
     resultMap(ingestedPreservica).s() should equal("true")
```
MancunianSam, line 535: There's a List on line 537 of the optional parameters which should be missing. Can you add digitalAssetSubtype to that list please.

MancunianSam (commented):
Looks good, couple of minor bits

scala/lambdas/dynamo-formatters/src/test/scala/uk/gov/nationalarchives/dynamoformatters/DynamoFormattersTest.scala
sparkhi: I've been doing some head scratching. This forall construct does not seem to fail the test even when one of the optional values is supplied. Something seem to be syntactically incorrect.
e.g. In this test, if I pass in some value `Option("digital asset subtype")` then the resultMap gets that value but the `resultMap.contains should be(false)` does not fail

MancunianSam (approved):

PR #308, opened 2024-10-23
DR2 1778 terrible time complexity of mapper lambda

MancunianSam:
I was getting the child count by iterating through every item in the
metadata.json and then counting every item in the same array with that
parent path so it was looping n^2 times which for a large n was taking
forever.

This makes a map of the parent path and count so we can lookup the
childCount.


scala/lambdas/ingest-mapper/src/main/scala/uk/gov/nationalarchives/ingestmapper/MetadataService.scala
```diff
@@ -88,7 +88,26 @@ class MetadataService(s3: DAS3Client[IO], discoveryService: DiscoveryService) {
...
+                val values = obj.value.toMap
+                val id = values("id").str
+                val parentPath = values.get("parentPath").map(parentPath => s"${parentPath.str}/").getOrElse("")
+                val objId = s"$parentPath$id"
+                val childCount = childCountMap.getOrElse(objId, 1)
+                Obj.from(values + ("childCount" -> Num(childCount)))
+              }
+
+              val departmentSeriesObjects = itemToDepartmentSeries.values.toList
+                .distinctBy(item => item.potentialSeriesItem.map(obj => obj("name")))
```
techncl, line 103: How come this is distinct by `name` and not `id`?

scala/lambdas/ingest-mapper/src/main/scala/uk/gov/nationalarchives/ingestmapper/MetadataService.scala
```diff
@@ -88,7 +88,26 @@ class MetadataService(s3: DAS3Client[IO], discoveryService: DiscoveryService) {
...
               val parentPaths = getParentPaths(json, itemToDepartmentSeries)
-              val updatedJson = jsonArr.map { metadataEntry =>
+              val childCountMap = parentPaths.groupBy(_._2).view.mapValues(_.size).toMap
+
+              def addChildCount(obj: Obj) = {
+                val values = obj.value.toMap
+                val id = values("id").str
+                val parentPath = values.get("parentPath").map(parentPath => s"${parentPath.str}/").getOrElse("")
+                val objId = s"$parentPath$id"
+                val childCount = childCountMap.getOrElse(objId, 1)
```
techncl, line 98: Is the reason why it's get or else 1 because we're dealing with folders that must have at least 1 child? If so, do you think:

1. It's worth adding a comment or changing the name of this method to `addChildCountToFolder`?
2. Adding a test for this?

scala/lambdas/ingest-mapper/src/main/scala/uk/gov/nationalarchives/ingestmapper/MetadataService.scala
```diff
@@ -88,7 +88,26 @@ class MetadataService(s3: DAS3Client[IO], discoveryService: DiscoveryService) {
           Stream.evals {
             topLevelIdsToDepartmentSeries.map { itemToDepartmentSeries =>
               val parentPaths = getParentPaths(json, itemToDepartmentSeries)
-              val updatedJson = jsonArr.map { metadataEntry =>
+              val childCountMap = parentPaths.groupBy(_._2).view.mapValues(_.size).toMap
+
+              def addChildCount(obj: Obj) = {
+                val values = obj.value.toMap
```
techncl, line 94: Maybe `entry`, `item` or `attributes` rather than `values`?

scala/lambdas/ingest-mapper/src/main/scala/uk/gov/nationalarchives/ingestmapper/MetadataService.scala
```diff
@@ -88,7 +88,26 @@ class MetadataService(s3: DAS3Client[IO], discoveryService: DiscoveryService) {
...
+                val id = values("id").str
+                val parentPath = values.get("parentPath").map(parentPath => s"${parentPath.str}/").getOrElse("")
+                val objId = s"$parentPath$id"
```
techncl, line 97: Maybe `pathToObj` or `objPath`?

scala/lambdas/ingest-mapper/src/main/scala/uk/gov/nationalarchives/ingestmapper/MetadataService.scala
```diff
@@ -88,7 +88,26 @@ class MetadataService(s3: DAS3Client[IO], discoveryService: DiscoveryService) {
           Stream.evals {
             topLevelIdsToDepartmentSeries.map { itemToDepartmentSeries =>
               val parentPaths = getParentPaths(json, itemToDepartmentSeries)
```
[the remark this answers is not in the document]
MancunianSam: It's `Map[UUID, String]` so `idToParent` makes sense. You know it isn't a full parent entity because it's a String. Again, I'd like to keep the variable names shorter and infer meaning from type signatures and context.

scala/lambdas/ingest-mapper/src/main/scala/uk/gov/nationalarchives/ingestmapper/MetadataService.scala
```diff
@@ -88,7 +88,26 @@ class MetadataService(s3: DAS3Client[IO], discoveryService: DiscoveryService) {
           Stream.evals {
             topLevelIdsToDepartmentSeries.map { itemToDepartmentSeries =>
```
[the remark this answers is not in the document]
MancunianSam: I think `idToDepartmentSeries` is ok as it's `UUID -> DepartmentSeriesTableItems`

scala/lambdas/ingest-mapper/src/main/scala/uk/gov/nationalarchives/ingestmapper/MetadataService.scala
```diff
@@ -88,7 +88,26 @@ class MetadataService(s3: DAS3Client[IO], discoveryService: DiscoveryService) {
...
               val parentPaths = getParentPaths(json, itemToDepartmentSeries)
-              val updatedJson = jsonArr.map { metadataEntry =>
+              val childCountMap = parentPaths.groupBy(_._2).view.mapValues(_.size).toMap
```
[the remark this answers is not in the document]
MancunianSam: It's `parentToChildCount` `To` is fine for `Map` but I rather not have too many variables with `And` and `Of` in it.

scala/lambdas/ingest-mapper/src/main/scala/uk/gov/nationalarchives/ingestmapper/MetadataService.scala
MancunianSam: The id is a random UUID so distinct by `id` won't deduplicate anything.

scala/lambdas/ingest-mapper/src/main/scala/uk/gov/nationalarchives/ingestmapper/MetadataService.scala
MancunianSam: This case is specifically for the department folder. That doesn't have a parent so isn't in the map. We're processing each top level department one at a time which means that each one will only have a single series. So if it's not in the map, it's a department folder which has 1 child.

I've added a comment to explain. The method name is fine as is and the `return the correct values from the lambda and upload the correct number of files and content to S3` is covering this path already.

scala/lambdas/ingest-mapper/src/main/scala/uk/gov/nationalarchives/ingestmapper/MetadataService.scala
MancunianSam: `attributes` is ok.

scala/lambdas/ingest-mapper/src/main/scala/uk/gov/nationalarchives/ingestmapper/MetadataService.scala
MancunianSam: This is the parent path of the children so I've called it `childParentPath`

techncl (commented):
👍

techncl (approved):

PR #314, opened 2024-10-25
DR2-1753 Support multiple inputs to allow batching

MancunianSam:


techncl (commented):
Looks good! Just a few comments.

1. Should we add tests for this to ensure that:
    1. we are filtering out `assetExists` in the asset opes creator
    2. the `existingAssets` and `missingAssets`in the `findExistingAsset` are correct
    3. the `updateSkipIngest` logic is correct in `findExistingAsset`
2. Would you mind updating the READMEs?

scala/lambdas/ingest-asset-opex-creator/src/main/scala/uk/gov/nationalarchives/ingestassetopexcreator/Lambda.scala
```diff
@@ -99,7 +104,9 @@ class Lambda extends LambdaRunner[Input, Unit, Config, Dependencies] {
 
 object Lambda {
 
-  case class Input(id: UUID, batchId: String, executionName: String)
+  case class InputItem(id: UUID, batchId: String, assetExists: Boolean)
```
techncl, line 107: Maybe `InputAsset`?

scala/lambdas/ingest-asset-opex-creator/src/main/scala/uk/gov/nationalarchives/ingestassetopexcreator/Lambda.scala
```diff
@@ -99,7 +104,9 @@ class Lambda extends LambdaRunner[Input, Unit, Config, Dependencies] {
...
+
+  case class Input(items: List[InputItem])
```
techncl, line 109: Maybe `assets` instead of `items`?

scala/lambdas/ingest-find-existing-asset/src/main/scala/uk/gov/nationalarchives/ingestfindexistingasset/Lambda.scala
```diff
@@ -64,9 +85,12 @@ class Lambda extends LambdaRunner[Input, StateOutput, Config, Dependencies] {
 object Lambda {
   case class Config(apiUrl: String, secretName: String, dynamoTableName: String) derives ConfigReader
 
-  case class Input(id: UUID, batchId: String)
+  case class InputItems(id: UUID, batchId: String)
+  case class OutputItems(id: UUID, batchId: String, assetExists: Boolean)
```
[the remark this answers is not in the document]
MancunianSam: 1. Don't know. Got confused. I've changed it to Item
2. Ok, the lambda is called find-existing-asset.
3. Yes it would :smile: We don't really care where it exists, just that it does. If we end up moving off a dedicated preservation system onto to central bucket storage, or putting it in DRI :cry:, or printing the bits on A4 paper or anything, we want the code to still make sense.

scala/lambdas/ingest-asset-opex-creator/src/main/scala/uk/gov/nationalarchives/ingestassetopexcreator/Lambda.scala
MancunianSam: Yep ok, it matches the other change now.

scala/lambdas/ingest-asset-opex-creator/src/main/scala/uk/gov/nationalarchives/ingestassetopexcreator/Lambda.scala
MancunianSam: We can't, that's what AWS provide and it's not configurable.

techncl (approved):
👍

PR #318, opened 2024-10-28
Increase role duration.

MancunianSam:
The default is 1 hour but the e2e tests take so long to run this isn't
enough. We'll try two hours for now and see how that goes.


techncl (approved):
👍

PR #321, opened 2024-10-29
Remove required property

MancunianSam:


techncl (commented):
👍 Sorry about that. Because one of our tickets was titled "digitalAssetSubtype should be null for TDR transfers", I thought `digitalAssetSubtype` would be required but can be `null`. So is that when `digitalAssetSubtype` is present, it can only be a String? If so, could you change the type on line 37 of the `asset-validation-schema.json` from `"type": ["string", "null" ]` to `"type": "string"`?

techncl (approved):
👍

PR #322, opened 2024-10-29
DR2-1973 Remove mockito from getPreservicaVersion lambda

MancunianSam:


techncl (commented):
Looks good! Nice to get away from mocking. 

I was just curious though, some of the tests that were removed, were ensuring that certain methods in the lambda were called so that if any were removed, the tests would fail or if one of the tests were removed, it would still check that the methods were called. Should we add something like that here? I'm thinking something similar to what was done in the Validation LambdaTest where we're using `Ref` to capture what was sent to the method?

MancunianSam:
No, I don't want to start down that route. There's an input to the lambda and initial state and there an output and final state. 

We can (and do in some of the more complicated lambdas) filter the test output based on the input to the methods. So here for example, instead of passing in `List[GetDr2PreservicaVersionResponse]` we could pass in `Map[String, List[GetDr2PreservicaVersionResponse]] where the string is a table. Then do something like:

```scala

    override def getItems[T, K](primaryKeys: List[K], tableName: String)(using returnFormat: DynamoFormat[T], keyFormat: DynamoFormat[K]): IO[List[T]] = IO.pure {
  primaryResponses.get(tableName).filter(res => matchesPrimaryKey(primaryKeys, res))
}
```
Or something like that so the test response is based on the inputs. 

The only reason I didn't bother here is because this is a really simple lambda and we'd have to really try to get the wrong inputs.

MancunianSam:
> There's an input to the lambda and **initial state** and there an output and **final state.**

It's the initial state turning to the final state that are the side effects, so initial and final SQS messages in a queue or initial S3 objects and final.

PR #338, opened 2024-11-07
DR2-1978 Remove mockito from court doc tests

MancunianSam:
I also got fed up of changing the json in the tar files in the test
resources so I've added in a method to generate a tar file in memory.

There's a bit of shuffling stuff around as well.


techncl (approved):
👍

techncl (approved):
👍

PR #340, opened 2024-11-07
DR2-1979 Remove mockito from start workflow.

MancunianSam:
This uses MurmurHash to generate an integer from the case class
argument. If you change the lambda code, it changes the hash which will
fail the test so we can be confident we're passing in the correct
arguments without checking them directly.


techncl (approved):
👍

PR #359, opened 2024-11-26
DR2-2004 Don't strip filenames and use description

MancunianSam:
We don't want to have the file extension stripped and we want to use the
description from TDR when available


kurtismash (approved):
Looks like code.

PR #360, opened 2024-11-27
Version bumps.

MancunianSam:
There were some version mismatches that were causing the secrets
rotation lambda to fail.

This should fix it.


kurtismash (approved):
👨‍💻

PR #365, opened 2024-11-28
DR2-1811 Alert when demo system upgrades

MancunianSam:
We should be sending slack messages via eventbridge. It's easiest
to just put it in the code.


techncl (approved):
👍

PR #370, opened 2024-12-04
DR2-1951 Opex creators assume role and write to bucket directly.

MancunianSam:
This is so we can get rid of datasync


techncl (approved):
👍

PR #395, opened 2025-01-16
Use ubuntu 22.04

MancunianSam:
The latest release has removed a lot of tools, one of which is sbt. This
is a temporary solution while we think of a clean way of installing sbt
on every build.


techncl (approved):
👍

PR #428, opened 2025-02-20
DR2-2117 Aggregator to return list of failures.

MancunianSam:
We need to do this
https://docs.aws.amazon.com/lambda/latest/dg/example_serverless_SQS_Lambda_batch_item_failures_section.html

This catches any execption, turns it into a custom case class with the
messageId from the message and then returns that as the result from the
lambda.

I've added some tests to check that we're returning the correct id.

### DIFF
--- a/scala/lambdas/preingest-tdr-aggregator/src/main/scala/uk/gov/nationalarchives/preingesttdraggregator/Lambda.scala
+++ b/scala/lambdas/preingest-tdr-aggregator/src/main/scala/uk/gov/nationalarchives/preingesttdraggregator/Lambda.scala
@@ -4,21 +4,22 @@ import cats.Monoid
 import cats.effect.IO
 import cats.effect.std.AtomicCell
 import cats.effect.unsafe.implicits.global
-import com.amazonaws.services.lambda.runtime.events.SQSEvent
+import com.amazonaws.services.lambda.runtime.events.{SQSBatchResponse, SQSEvent}
 import com.amazonaws.services.lambda.runtime.{Context, RequestHandler}
 import io.circe.generic.auto.*
 import pureconfig.ConfigReader.Result
 import pureconfig.module.catseffect.syntax.*
 import pureconfig.{ConfigCursor, ConfigReader, ConfigSource}
+import uk.gov.nationalarchives.preingesttdraggregator.Duration.*
 import uk.gov.nationalarchives.preingesttdraggregator.Ids.GroupId
 import uk.gov.nationalarchives.preingesttdraggregator.Lambda.*
 import uk.gov.nationalarchives.{DADynamoDBClient, DASFNClient}
-import uk.gov.nationalarchives.preingesttdraggregator.Duration.*
+
 import java.net.URI
 import java.time.Instant
 import java.util.UUID
 import scala.jdk.CollectionConverters.*
-class Lambda extends RequestHandler[SQSEvent, Unit]:
+class Lambda extends RequestHandler[SQSEvent, SQSBatchResponse]:
 
   given Monoid[Map[String, Group]] = Monoid.instance(Map.empty, _ ++ _)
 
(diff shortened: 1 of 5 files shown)


scala/lambdas/preingest-tdr-aggregator/src/main/scala/uk/gov/nationalarchives/preingesttdraggregator/Lambda.scala
```diff
@@ -28,13 +29,12 @@ class Lambda extends RequestHandler[SQSEvent, Unit]:
...
-    ConfigSource.default
-      .loadF[IO, Config]()
-      .flatMap { config =>
-        Aggregator[IO].aggregate(config, groupCacheAtomicCell, input.getRecords.asScala.toList, context.getRemainingTimeInMillis)
-      }
-      .unsafeRunSync()
+  override def handleRequest(input: SQSEvent, context: Context): SQSBatchResponse = {
+    for {
+      config <- ConfigSource.default.loadF[IO, Config]()
+      results <- Aggregator[IO].aggregate(config, groupCacheAtomicCell, input.getRecords.asScala.toList, context.getRemainingTimeInMillis)
```
techncl, line 35: Maybe this could be "potentialFailures" or "BatchItemFailures" or something?

scala/lambdas/preingest-tdr-aggregator/src/main/scala/uk/gov/nationalarchives/preingesttdraggregator/Lambda.scala
```diff
@@ -28,13 +29,12 @@ class Lambda extends RequestHandler[SQSEvent, Unit]:
...
+    } yield SQSBatchResponse.builder().withBatchItemFailures(results.asJava).build
```
techncl, line 36: What happens if there are no failures, does just send a failure message with nothing in it?

techncl (commented):
Looks good! Just some minor comments, as well as:

1. Now that we're returning something from the Lambda, should we add tests for it?
2. Could you update the README please?

scala/lambdas/preingest-tdr-aggregator/src/main/scala/uk/gov/nationalarchives/preingesttdraggregator/Lambda.scala
MancunianSam: `potentialFailures` is good because there might be none.

scala/lambdas/preingest-tdr-aggregator/src/main/scala/uk/gov/nationalarchives/preingesttdraggregator/Lambda.scala
MancunianSam: Yep, it sends 
```json
{ 
  "batchItemFailures": []
}
```

MancunianSam:
1. I don't think so. The lambda does very little and the tests would be almost identical to the aggregator ones.
2. I've updated the README.

scala/lambdas/preingest-tdr-aggregator/src/main/scala/uk/gov/nationalarchives/preingesttdraggregator/Lambda.scala
techncl: Should we prevent it from sending anything if there are no failures or does it not matter?

techncl:
> 1. I don't think so. The lambda does very little and the tests would be almost identical to the aggregator ones.

Ah, no, I meant like since `handleRequest` returns an `SQSBatchResponse`, should we have maybe 1 or 2 tests that call `handleRequest` (with a mocked `ConfigSource.default.loadF[IO, Config]()` and `Aggregator[IO].aggregate` response) in order to ensure that the Lambda is returning a `SQSBatchResponse` (We're sure that the Aggregator is but not the Lambda) and also that it is passing in the correct arguments to the `aggregate` method?

scala/lambdas/preingest-tdr-aggregator/src/main/scala/uk/gov/nationalarchives/preingesttdraggregator/Lambda.scala
MancunianSam: You have to send the batch response in that format if we've configured the lambda to expect it.

MancunianSam:
I've added in a couple of lambda tests and I've refactored it a bit to make it less implicit-y

techncl (approved):
👍

PR #429, opened 2025-02-21
DR2-2097 Update metadata validation.

MancunianSam:
The metadata validation in the case class in the aggreagator

```scala
  case class TDRMetadata(
      Series: String,
      UUID: UUID,
      description: Option[String],
      TransferringBody: String,
      TransferInitiatedDatetime: String,
      ConsignmentReference: String,
      Filename: String,
      SHA256ServerSideChecksum: String,
      FileReference: String
  )
```

is more strict than the validation here which means things are getting
to the aggregator which are going to fail which is a waste of time.

This updates the metadata schema to match and adds tests to check for
the changes.


kurtismash:
Part of the original ticket to implement this validation was to only validate the fields that would cause a failure further downstream; as long as the fields we require as present we should continue to ingest.

Where fields have been defined that aren't required will this implement additional protection? E.g. What happens if `description` comes in as a number?

MancunianSam:
@techncl
1. Good idea, I've added one in and removed the existing ones .
2. No, we don't care.
3. Yes. While I was doing this, I found that it wasn't validating the uuid format so I've changed the validator so it does now. 
4. Yeah we should. I've moved the schema out of the file and added a test in the package builder to create a json string from the case class and validate it. There is a problem in that the date in the case class is a `String` because the TDR export isn't sending a properly formatted date. I tried to work around it but it'll probably make things worse so I'll just fix the export at some point. 

@kurtismash 
I'm not sure what you're asking. These are fields we need further downstream as far as I know. If description comes in as a number then the validation will fail. Or maybe we don't want it to, I'm not sure.

kurtismash:
@MancunianSam I guess it's a question of "What would Scala do" when parsing the JSON. I ask because we're not requiring a `description` but are requiring it be a string if there is one, whereas we passthrough every other non-required field.

MancunianSam:
@kurtismash That is what Scala would do. The case class in the first comment on this PR has description as an Option, so it doesn't have to be there but needs to a string if it is.
The rest are mandatory and UUID needs to be a UUID.

python/lambdas/copy-files-from-tdr/copy_files_lambda_function_test.py
```diff
@@ -156,7 +156,8 @@ def test_send_message_failure(self, mock_validate_metadata, mock_send_message, _
 
     def test_should_successfully_validate_when_the_fields_are_valid(self):
         mock_response_body = json.dumps(self.valid_metadata())
-        result = lambda_function.validate_mandatory_fields_exist(json.loads(mock_response_body))
+        schema_location = "python/lambdas/copy-files-from-tdr/metadata-schema.json"
```
techncl, line 159: Should this be a global variable?

python/lambdas/copy-files-from-tdr/copy_files_lambda_function_test.py
```diff
@@ -203,7 +204,8 @@ def selective_error_object(Bucket, Key):
             str(ex.exception))
 
     def test_should_raise_an_exception_if_fields_are_missing(self):
-        with open("metadata-schema.json", "r") as metadata_schema_file:
+        schema_location = "python/lambdas/copy-files-from-tdr/metadata-schema.json"
+        with open(schema_location, "r") as metadata_schema_file:
```
techncl, line 208: Should this be global since we are doing it twice?

python/lambdas/copy-files-from-tdr/copy_files_lambda_function_test.py
MancunianSam: I'm going to leave it. Too many global variables in tests makes them hard to read as you're jumping around everywhere to see what it's doing. These tests aren't long so it's fine.

python/lambdas/copy-files-from-tdr/copy_files_lambda_function_test.py
MancunianSam: As above.

scala/lambdas/preingest-tdr-package-builder/src/test/scala/uk/gov/nationalarchives/preingesttdrpackagebuilder/MetadataSchemaTest.scala
```diff
@@ -0,0 +1,50 @@
...
+import java.time.{Instant, LocalDateTime, ZoneId}
+import java.util.UUID
+
+class MetadataSchemaTest extends AnyFlatSpec {
+
+  "the metadata schema" should "validate against the TDRMetadata case class" in {
+    val projectRoot = new File(".").getCanonicalPath
+      .replace("/scala/lambdas/preingest-tdr-package-builder", "")
+
+    val schemaPath = s"$projectRoot/python/lambdas/copy-files-from-tdr/metadata-schema.json"
```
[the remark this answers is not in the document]
MancunianSam: We shouldn't forget because the tests will fail if they get out of sync but I'm happy to move it. Should we put it in the root of this repo or somewhere else?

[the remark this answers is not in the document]
kurtismash: Maybe a top level `common/preingest-tdr` path? I'm sure they'll be more things we want to share between Lambdas, and I suspect we're doing some configuration stuff like this through Terraform already.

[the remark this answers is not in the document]
MancunianSam: Ok makes sense. I've moved it there and updated the deploy script to put it into the right place.

kurtismash (approved):

PR #441, opened 2025-03-12
DR2-2130 Only look at events 10 minutes old.

MancunianSam:
This will now:
* Get the updated since date from Dynamo
* Get entities-updated-since from Preservica for events 10 minutes
  before that time.
* Get the event actions for all entities and get the latest action time.
* If the latest time is before the event action time - 10 minutes, send
  the message.


kurtismash:
> Get entities-updated-since from Preservica for events 10 minutes
before that time.

I don't think this is the intended effect. My understanding is that the value in the state table is the datetime of the latest event we have already sent a notification for. With this change we would want to continue the current logic of finding entities that have been updated AFTER the datetime from DynamoDB but, now, before `now() - 10 mins`. And continue to protect against the race condition where an entity is updated between our calls to `/updated-since` and `/event-actions`.

The API equivalent of `SELECT entity ... WHERE updatedTime BETWEEN stateDatetime AND NOW() - INTERVAL '10 minutes'`.

MancunianSam:
Ok, makes sense I think. I've changed it so the only time we're taking 10 minutes off is with the current time so it's now
* Get events after Dynamo time
* Get the latest event actions for these
* Get the most recent.
* Check if this is before (currentTime - 10 mins)

techncl (approved):
👍

PR #448, opened 2025-03-17
Update vulnerable dependencies.

MancunianSam:
There was an old secrets manager sdk that we're not actually using so
I've removed that.

Also, scanamo is using an old version of the Dynamo library which might
also be a problem so I've overridden that.


techncl (approved):
👍

PR #454, opened 2025-03-19
Add json output on error.

MancunianSam:
This makes it easier fo the e2e tests to find errors and it should give
us something to search for in Cloudwatch as well.

This is a quick fix to get the tests working. A better logging solution will be implemented later

### DIFF
--- a/python/lambdas/copy-files-from-tdr/lambda_function.py
+++ b/python/lambdas/copy-files-from-tdr/lambda_function.py
@@ -20,15 +20,19 @@ def lambda_handler(event, context):
         file_id = body["fileId"]
         metadata_file_id = f"{file_id}.metadata"
         source_bucket = body["bucket"]
-        files = [file_id, metadata_file_id]
-        assert_objects_exist_in_bucket(source_bucket, files)
-        validate_metadata(source_bucket, metadata_file_id)
-        file_location = copy_objects(destination_bucket, file_id, source_bucket)
-        copy_objects(destination_bucket, metadata_file_id, source_bucket)
-        potential_message_id = {key: value for key, value in body.items() if key == "messageId"}
-        sqs_body = {"id": file_id, "location": file_location}
-        sqs_body.update(potential_message_id)
-        sqs_client.send_message(QueueUrl=destination_queue, MessageBody=json.dumps(sqs_body))
+        try:
+            files = [file_id, metadata_file_id]
+            assert_objects_exist_in_bucket(source_bucket, files)
+            validate_metadata(source_bucket, metadata_file_id)
+            file_location = copy_objects(destination_bucket, file_id, source_bucket)
+            copy_objects(destination_bucket, metadata_file_id, source_bucket)
+            potential_message_id = {key: value for key, value in body.items() if key == "messageId"}
+            sqs_body = {"id": file_id, "location": file_location}
+            sqs_body.update(potential_message_id)
+            sqs_client.send_message(QueueUrl=destination_queue, MessageBody=json.dumps(sqs_body))
+        except Exception as e:
+            print(json.dumps({"error": str(e), "fileId": file_id}))
+            raise Exception(str(e))
 
 
 def assert_objects_exist_in_bucket(source_bucket, files):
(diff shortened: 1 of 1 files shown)


python/lambdas/copy-files-from-tdr/lambda_function.py
```diff
@@ -20,15 +20,19 @@ def lambda_handler(event, context):
...
-        assert_objects_exist_in_bucket(source_bucket, files)
-        validate_metadata(source_bucket, metadata_file_id)
-        file_location = copy_objects(destination_bucket, file_id, source_bucket)
-        copy_objects(destination_bucket, metadata_file_id, source_bucket)
-        potential_message_id = {key: value for key, value in body.items() if key == "messageId"}
-        sqs_body = {"id": file_id, "location": file_location}
-        sqs_body.update(potential_message_id)
-        sqs_client.send_message(QueueUrl=destination_queue, MessageBody=json.dumps(sqs_body))
+        try:
+            files = [file_id, metadata_file_id]
```
techncl, line 24: Does this line need to be in the `try`?

python/lambdas/copy-files-from-tdr/lambda_function.py
```diff
@@ -20,15 +20,19 @@ def lambda_handler(event, context):
...
+            assert_objects_exist_in_bucket(source_bucket, files)
+            validate_metadata(source_bucket, metadata_file_id)
+            file_location = copy_objects(destination_bucket, file_id, source_bucket)
+            copy_objects(destination_bucket, metadata_file_id, source_bucket)
```
techncl, line 28: Do you think it's worth splitting this `try` into two `try`s: one, from 24-28 for s3 errors and then line 29-32 for SQS, and then we could change the error name to be "S3 Error" or "SQS Error" and maybe incorporate more domain specific information in each error such as the destination_bucket or destination_queue?

python/lambdas/copy-files-from-tdr/lambda_function.py
```diff
@@ -20,15 +20,19 @@ def lambda_handler(event, context):
...
+            potential_message_id = {key: value for key, value in body.items() if key == "messageId"}
+            sqs_body = {"id": file_id, "location": file_location}
+            sqs_body.update(potential_message_id)
+            sqs_client.send_message(QueueUrl=destination_queue, MessageBody=json.dumps(sqs_body))
+        except Exception as e:
+            print(json.dumps({"error": str(e), "fileId": file_id}))
```
techncl, line 34: Should we capitalise the "error" to make it easier to spot?

python/lambdas/copy-files-from-tdr/lambda_function.py
```diff
@@ -20,15 +20,19 @@ def lambda_handler(event, context):
...
+            raise Exception(str(e))
```
techncl, line 35: I don't think you need to convert the exception to a string

techncl (commented):
Looks good! Do you think it's worth adding testing for this?

python/lambdas/copy-files-from-tdr/lambda_function.py
MancunianSam: No, I've moved it.

python/lambdas/copy-files-from-tdr/lambda_function.py
MancunianSam: It's a good idea but `file_location` is needed in the calls to SQS so we'd need nested try blocks which is a bit odd. 
Also, we're already printing `f"Error during copy of '{s3_key}' from '{source_bucket}' to '{destination_bucket}': {e}"` in the copy_objects function if there was an error.

This is meant as a bit of a catch all "This failed" message and then why it's failed can be looked at elsewhere.

python/lambdas/copy-files-from-tdr/lambda_function.py
MancunianSam: No, it's json, camel case all the way.

python/lambdas/copy-files-from-tdr/lambda_function.py
MancunianSam: Oh yeah thanks. I was getting test failures and I thought this would help but it didn't and I forgot to put it back.

MancunianSam:
It's going to be tricky to add a test for a print statement. The tests that are expecting exceptions still pass so we're not supressing anything we shouldn't be.

techncl (commented):
👍

PR #483, opened 2025-04-29
Get ingest timings for report

techncl:


python/lambdas/ingest-sfn-timings/README.md
```diff
@@ -0,0 +1,48 @@
...
+   ```
+
+
+## Environment Variables
+
+| Name               | Description                                                                               |
+|--------------------|-------------------------------------------------------------------------------------------|
+| INGEST_SFN_ARN     | The state machine arn of the ingest step function                                         |
+| WORKFLOW_SFN_ARN   | The state machine arn of the workflow step function that the ingest step function calls   |
+| PREINGEST_SFN_ARN  | The execution arn of the preingest step function (that triggers the ingest step function) |
```
kurtismash, line 45: What do we do when there's more than 1 preingest Step Function? We already have 2 preingest stacks - COURTDOC and TDR.

python/lambdas/ingest-sfn-timings/README.md
```diff
@@ -0,0 +1,48 @@
...
+   ```json
+                {
+                    "name": "ingestExecutionName",
+                    "steps": {
+                         "preingest": {
+                            "endDate": 1735776000.0,
+                            "name": "preingestExecutionName",
+                            "startDate": 1735776000.0
+                        },
+                         "preservicaIngest": {
```
kurtismash, line 24: Please can we not use the word "preservica", DR2 should be preservation system agnostic.

python/lambdas/ingest-sfn-timings/sfn_timings/lambda_function.py
```diff
@@ -0,0 +1,118 @@
+import json
+import os
+from datetime import datetime, timedelta
+
+import boto3
+from pytz import UTC
+
+sfn_client = boto3.client("stepfunctions", region_name="eu-west-2")
+s3_client = boto3.client("s3", region_name="eu-west-2")
```
kurtismash, line 9: Is the region name required here? The SDK should pull it from the `AWS_REGION` environment variable if not passed explicitly. By including here we're restricting this Lambda function to only work in eu-west-2.

python/lambdas/ingest-sfn-timings/sfn_timings/lambda_function.py
```diff
@@ -0,0 +1,118 @@
...
+common_exec_args = {"maxResults": 100, "statusFilter": "SUCCEEDED"}
+
+
+def next_token(executions_info):
+    token = executions_info.get("nextToken", "")
+    return {} if token == "" else {"nextToken": token}
+
+
+def get_workflow_executions(cut_off_date, workflow_sfn_arn, relevant_executions=[], next_page_token={}) \
```
kurtismash, line 18: There's a Python gotcha here, `relevant_executions` and `next_page_token` will not be reset to an empty list/dict on subsequent calls (e.g. reusing the same Lambda container) - https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments

python/lambdas/ingest-sfn-timings/sfn_timings/lambda_function.py
```diff
@@ -0,0 +1,118 @@
...
+from pytz import UTC
+
+sfn_client = boto3.client("stepfunctions", region_name="eu-west-2")
+s3_client = boto3.client("s3", region_name="eu-west-2")
+common_exec_args = {"maxResults": 100, "statusFilter": "SUCCEEDED"}
+
+
+def next_token(executions_info):
+    token = executions_info.get("nextToken", "")
+    return {} if token == "" else {"nextToken": token}
```
kurtismash, line 15: Would using the built in boto3 paginator be easier?

python/lambdas/ingest-sfn-timings/sfn_timings/lambda_function.py
```diff
@@ -0,0 +1,118 @@
...
+                                        preingest_sfn_arn)
+
+    for ingest_timing in ingest_timings:
+        json_file_key = f"""step-function-timings-for-{ingest_timing["name"]}.json"""
+        ingest_timing_json_obj = json.dumps(ingest_timing)
+
+        try:
+            s3_client.put_object(Body=ingest_timing_json_obj, Bucket=bucket_name, Key=json_file_key, ContentType="application/json")
+        except Exception as e:
+            print(f"There was an error when attempting to upload file {json_file_key} to bucket {bucket_name}: {e}")
```
kurtismash, line 118: What's the benefit of catching the error here?

python/lambdas/ingest-sfn-timings/sfn_timings/lambda_function.py
```diff
@@ -0,0 +1,118 @@
...
+        else:
+            next_page_token = next_token(ingest_sfn_executions_info)
+            executions = get_ingest_timings(cut_off_date, workflow_sfn_executions, ingest_sfn_arn,
+                                            preingest_sfn_arn, executions, next_page_token)
+
+    return executions
+
+
+def lambda_handler(event, context):
+    event_triggered_date = UTC.localize(datetime.strptime(event["time"], "%Y-%m-%dT%H:%M:%SZ"))
```
kurtismash, line 98: Is UTC required here? Is the time generated by EventBridge always in UTC already?

python/lambdas/ingest-sfn-timings/sfn_timings/lambda_function.py
```diff
@@ -0,0 +1,118 @@
...
+                    "preservicaIngest": {
+                        "name": name,
+                        "startDate": execution["startDate"].timestamp(),
+                        "endDate": execution["stopDate"].timestamp()
+                    }
+                }
+            }
+
+            if name.startswith("TDR_"):
+                preingest_sfn_execution_info = sfn_client.describe_execution(executionArn=f"{preingest_sfn_arn}:{name}")
```
kurtismash, line 70: What happens if an execution of the other Step Functions doesn't exist?

python/lambdas/ingest-sfn-timings/sfn_timings/lambda_function.py
```diff
@@ -0,0 +1,118 @@
...
+
+                execution_info["steps"]["preingest"] = {
+                    "name": preingest_sfn_execution_info["name"],
+                    "startDate": preingest_sfn_execution_info["startDate"].timestamp(),
+                    "endDate": preingest_sfn_execution_info["stopDate"].timestamp()
+                }
+
+            workflow_execution = \
+            [workflow_ex for workflow_ex in workflow_sfn_executions if name in workflow_ex["name"]][0]
```
kurtismash, line 79: Why do we fetch the details of the dr2-preingest-tdr execution within this function but prefetch the dr2-ingest-run-workflow executions?

python/lambdas/ingest-sfn-timings/sfn_timings/lambda_function.py
```diff
@@ -0,0 +1,118 @@
...
+            executions = get_ingest_timings(cut_off_date, workflow_sfn_executions, ingest_sfn_arn,
+                                            preingest_sfn_arn, executions, next_page_token)
+
+    return executions
+
+
+def lambda_handler(event, context):
+    event_triggered_date = UTC.localize(datetime.strptime(event["time"], "%Y-%m-%dT%H:%M:%SZ"))
+    last_ingest_completion_date = event_triggered_date - timedelta(
+        minutes=10)  # 10 minutes since the last lambda execution
```
kurtismash, line 100: AWS EventBridge only guarantees that events will be sent within the minute they are scheduled, in theory there could be over 10 minutes between invocations.

kurtismash (commented):
Hi @techncl, this is a good start but I'm confused by some of the choices made here and the general flow of the application. I fear you've tried to be too Scala-ry for Python. **I have not reviewed the tests file.**

Moreover, the AWS architecture this fits into concerns me. The function is to be invoked on a schedule via EventBridge and lists the executions of our Step Functions since it last ran, filtering only for successfully completed executions. Would it be better to invoke the function within the Step Function or triggered by the Step Function Status Change event? The function generates JSON with the timings of the 3 Step Function executions, what will this be used for - could it's use case parse it from existing Step Function or DR2 notification logs? When the file is in S3, how is it's lifecycle to be managed to enable time based reporting?

python/lambdas/ingest-sfn-timings/sfn_timings/lambda_function.py
techncl: I'm not sure what you mean. Do you mean if a TDR ingest step function doesn't have a preingest one preceding (triggering) it? If so, how?

python/lambdas/ingest-sfn-timings/sfn_timings/lambda_function.py
techncl: Because of the way the workflow step function was named, it starts with the step function name but then has an execution id which I couldn't find anywhere

python/lambdas/ingest-sfn-timings/sfn_timings/lambda_function.py
techncl: Just adding more information. because I wasn't sure if all of the errors that could be thrown when uploading the file would contain the file name/bucket and thought that might be useful information to have especially if only one item/few items fails to upload

python/lambdas/ingest-sfn-timings/sfn_timings/lambda_function.py
techncl: so is the event "time" the date it actually triggered or when it was supposed to be triggered? If it is the latter then it should be fine, right, since it's consistent?

techncl:
> I fear you've tried to be too Scala-ry for Python. **I have not reviewed the tests file.**
How so, because of the recursion?

> Moreover, the AWS architecture this fits into concerns me. The function is to be invoked on a schedule via EventBridge and lists the executions of our Step Functions since it last ran, filtering only for successfully completed executions. Would it be better to invoke the function within the Step Function or triggered by the Step Function Status Change event? The function generates JSON with the timings of the 3 Step Function executions, what will this be used for - could it's use case parse it from existing Step Function or DR2 notification logs? When the file is in S3, how is it's lifecycle to be managed to enable time based reporting?

What you are saying makes sense (as it would reduce the possibility of missing a lambda if the timings are milliseconds off) but it might be best to ask @MancunianSam about these things since I remember you bringing these concerns up in a meeting and I think the conclusion was to just do this as it's a proof of concept

python/lambdas/ingest-sfn-timings/sfn_timings/lambda_function.py
kurtismash: You're correct that the error from the SDK won't include the bucket/key, which makes it near impossible to debug. But as you then don't raise the exception again the Lambda will complete successfully. If there's an intermittent issue the Lambda wouldn't retry in this case.

python/lambdas/ingest-sfn-timings/sfn_timings/lambda_function.py
kurtismash: Say you have an event triggered at 00:00:00 but then the next event isn't triggered until 00:10:59, there's 59 second unaccounted for. Unless the time in the event is always to the minute 🤔.

python/lambdas/ingest-sfn-timings/sfn_timings/lambda_function.py
techncl: Ah, I see what you mean, I'll rethrow the error

MancunianSam (dismissed):

MancunianSam (approved):

PR #486, opened 2025-05-02
Update Mapper Lambda Output with groupId

techncl:
Because the failure notifications lambda needs the 'groupId', the workflow step function needs it and because the workflow step function needs it, the Mapper should return it so that it is passed along. The Lambda currently bases the groupId on the batchId but this is only until the step function starts passing it in on its own


MancunianSam (approved):
:+1:

PR #511, opened 2025-05-21
Add io ref to reconciler output.

MancunianSam:
We need to add the ioRef to the output so that this is added to the
post-ingest table so that the CC confirmer can look this up in the OCFL
repo without having the confirmer contact Preservica.


techncl (approved):
👍

PR #525, opened 2025-06-05
Dr2 2259 add post ingest change handler

techncl:


scala/lambdas/postingest-state-change-handler/src/main/scala/uk/gov/nationalarchives/postingeststatechangehandler/Lambda.scala
```diff
@@ -0,0 +1,213 @@
...
+
+  case class Config(dynamoTableName: String, dynamoGsiName: String, topicArn: String, queues: String) derives ConfigReader
+
+  case class DynamodbEvent(Records: List[DynamodbStreamRecord])
+
+  case class DynamodbStreamRecord(eventName: EventName, dynamodb: StreamRecord)
+
+  case class StreamRecord(keys: Option[FilesTablePrimaryKey], oldImage: Option[PostIngestStatusTableItem], newImage: PostIngestStatusTableItem)
+
+  case class Queue(queueAlias: String, queueOrder: Int, queueUrl: String) {
```
MancunianSam, line 208: I'm not sure about this dynamic lookup of attributes. This will fail if we change the attribute names although the tests would catch it. I tried something like this
```scala
given Decoder[Queue] = (c: HCursor) => for {
    queueAlias <- c.downField("queueAlias").as[String]
    queueOrder <- c.downField("queueOrder").as[Int]
    queueUrl <- c.downField("queueUrl").as[String]
  } yield queueAlias match 
    case "CC" => CCQueue(queueAlias, queueOrder, queueUrl)


  sealed trait Queue {
    def queueAlias: String
    def queueOrder: Int
    def queueUrl: String
    def resultAttrName: String
    def resultFn: PostIngestStatusTableItem => Option[String]
  }

  case class CCQueue(queueAlias: String, queueOrder: Int, queueUrl: String) extends Queue {
    val resultAttrName: String = "CC"
    val resultFn: PostIngestStatusTableItem => Option[String] = tableItem => tableItem.potentialResultCC
  }
```

Then the getModifyFibers method will just have
```scala
val potentialQueue = queues.find { queue =>
                  queue.resultFn(newItem) != queue.resultFn(oldItem)
                }
```

What do you think?

scala/lambdas/postingest-state-change-handler/src/main/scala/uk/gov/nationalarchives/postingeststatechangehandler/Lambda.scala
```diff
@@ -0,0 +1,213 @@
...
+
+    def getInsertFibers(queues: List[Queue]) = event.Records
+      .filter(_.eventName == EventName.INSERT)
+      .parTraverse { record =>
+        val queue1 = queues.find(_.queueOrder == 1).get
+        val processInsertRecord = updateTableAndSendToSqs(record.dynamodb.newImage, queue1) >> sendOutputMessage(record.dynamodb.newImage, Some(queue1.queueAlias))
+        processInsertRecord.start
+      }
+
+    def newResultDiffersFromOld(newResult: Option[String], oldResult: Option[String]) = newResult.getOrElse("") != oldResult.getOrElse("")
```
MancunianSam, line 82: You don't need to do getOrElse, if you compare `Option` directly then it works and if you do that you don't really need a seperate method either.

scala/lambdas/postingest-state-change-handler/src/test/scala/uk/gov/nationalarchives/postingeststatechangehandler/LambdaTest.scala
```diff
@@ -0,0 +1,172 @@
...
+    sentSnsMessage.properties.parentMessageId should equal(Some(correlationId))
+    sentSnsMessage.properties.timestamp should equal(Instant.parse("2038-01-19T03:14:07Z"))
+    sentSnsMessage.properties.messageType should equal(IngestUpdate)
+
+    sentSnsMessage.parameters.assetId should equal(fullyPostIngestedAsset.assetId)
+    sentSnsMessage.parameters.status should equal(IngestedPreservation)
+  }
+
+  forAll(updateTableScenarios) { (oldImage, newImage, description) =>
+    "handler" should f"delete the item if the event is an 'MODIFY' one, $description" in {
```
MancunianSam, line 105: I think you can just use `s""` here, you don't need `f""`

MancunianSam (commented):
Looks good. Just a few bits and a suggestion.

scala/lambdas/postingest-state-change-handler/src/test/scala/uk/gov/nationalarchives/postingeststatechangehandler/LambdaTest.scala
techncl: 👍 😄 That was a mistake; I think I had a slight case of Python-brain when I did this

scala/lambdas/postingest-state-change-handler/src/main/scala/uk/gov/nationalarchives/postingeststatechangehandler/Lambda.scala
techncl: Wait! Both of them can be `None` if for example a different attribute gets changed e.g.

```
// oldImage is this
PostIngestStatusTableItem(1, "batchId", "input", Some("correlationId"), Some(queue1), "dateTimeYesterday", "dateTimeYesterday", potentialResultCC=None)
```
and 
```

// newImage is this
PostIngestStatusTableItem(1, "batchId", "input", Some("correlationId"), Some(queue1), "dateTimeToday", "dateTimeToday", potentialResultCC=None)
```

This would trigger a MODIFY event, right?

And then when you do a `!newResult.contains(oldResult)` (i.e `!newPotentialResultCC.contains(oldPotentialResultCC)`), you would get `true`

scala/lambdas/postingest-state-change-handler/src/main/scala/uk/gov/nationalarchives/postingeststatechangehandler/Lambda.scala
techncl: I like this; it's like the midway between my original attempt and my current attempt

kurtismash: Seeing we have a class `CCQueue`, would this still work if we added another queue? Ideally, this Lambda should run dynamically - given a structure of queues as an environment variable the Lambda should route each item through our linear, sequential pipeline without knowing anything more than the information provided in the environment variable.

MancunianSam (approved):
:+1:

PR #536, opened 2025-06-10
Use dependency review action

MancunianSam:
There are two steps.
1. Use sbt-dependency-submission to create the dependency graph
2. Use dependency-review-action to scan for vulnerabilities

I've replaced the dr2-github-actions workflow as this is the only project still using it.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -10,8 +10,8 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - uses: sbt/setup-sbt@v1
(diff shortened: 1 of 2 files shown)


techncl (commented):
Looks good! Some minor comments.

Just out of curiosity, now that you've changed it to push, does that mean that the tests won't run on every push for our commits? If so, how come?

paulschwarzenberger:
suggest pinning actions to commit hashes:
```
- uses: scalacenter/sbt-dependency-submission@v3
- uses: actions/dependency-review-action@v4
```

PR #544, opened 2025-06-16
DR2-2128 Fix entity event generator

MancunianSam:
The problem, in pseudo-ish code is:

```
def run(start):
    total_sent = send_messages(start)
    if total_sent > 0:
        run(1000)

def send_messages(start):
    date = get_date_from_dynamo()
    entities = entities_updated_since(date, start)
    last_updated = get_last_updated_from_event_actions()
    update_dynamo_with_last_updated()
    return length(entities)

run(0)
```

So for a given run we would call `entities_updated_since(2025-06-10T10:00, 0)`
This would yield an last updated time from the event actions of 2025-06-12T09:00
Dynamo would be updated with 2025-06-12T09:00
The method returns 1000.
The method calls itself with start + 1000
The next call to entities_updated_since is with 2025-06-12T09:00 and 1000
We're missing the first 1000 rows for that time.

This isn't a problem if there are fewer than 1000 rows since the last
updated time but this isn't always true.

The fix is to store the start number in Dynamo as well. This is almost
always 0. The exception is if there are more than 1000 things with the
exact same last updated time (unlikely but not impossible)

I've run this and we're now seeing all the things in CC that we're
expecting.

### DIFF
--- a/scala/lambdas/entity-event-generator-lambda/src/main/scala/uk/gov/nationalarchives/entityeventgenerator/Lambda.scala
+++ b/scala/lambdas/entity-event-generator-lambda/src/main/scala/uk/gov/nationalarchives/entityeventgenerator/Lambda.scala
@@ -8,18 +8,17 @@ import pureconfig.ConfigReader
 import software.amazon.awssdk.services.dynamodb.model.*
 import sttp.capabilities.fs2.Fs2Streams
 import uk.gov.nationalarchives.DADynamoDBClient.DADynamoDbRequest
-import uk.gov.nationalarchives.{DADynamoDBClient, DASNSClient}
-import uk.gov.nationalarchives.utils.EventCodecs.given
-import uk.gov.nationalarchives.utils.LambdaRunner
-import uk.gov.nationalarchives.entityeventgenerator.Lambda.*
 import uk.gov.nationalarchives.dp.client.Entities.Entity
 import uk.gov.nationalarchives.dp.client.EntityClient
 import uk.gov.nationalarchives.dp.client.fs2.Fs2Client
+import uk.gov.nationalarchives.entityeventgenerator.Lambda.*
+import uk.gov.nationalarchives.utils.EventCodecs.given
+import uk.gov.nationalarchives.utils.LambdaRunner
+import uk.gov.nationalarchives.{DADynamoDBClient, DASNSClient}
 
 import java.time.{Instant, OffsetDateTime}
 
 class Lambda extends LambdaRunner[ScheduledEvent, Int, Config, Dependencies] {
-  private val maxEntitiesPerPage: Int = 1000
   private val dateItemPrimaryKeyAndValue =
     Map("id" -> AttributeValue.builder().s("LastPolled").build())
   private val datetimeField = "datetime"
@@ -32,7 +31,6 @@ class Lambda extends LambdaRunner[ScheduledEvent, Int, Config, Dependencies] {
       entityClient: EntityClient[IO, Fs2Streams[IO]],
       dADynamoDBClient: DADynamoDBClient[IO],
       dASnsDBClient: DASNSClient[IO],
-      startFrom: Int,
       eventTriggeredDatetime: OffsetDateTime
   ): IO[Int] =
     for {
@@ -41,7 +39,6 @@ class Lambda extends LambdaRunner[ScheduledEvent, Int, Config, Dependencies] {
         entityClient,
         dADynamoDBClient,
         dASnsDBClient,
-        startFrom,
         eventTriggeredDatetime
       )
       _ <-
@@ -51,7 +48,6 @@ class Lambda extends LambdaRunner[ScheduledEvent, Int, Config, Dependencies] {
             entityClient,
             dADynamoDBClient,
             dASnsDBClient,
-            startFrom + maxEntitiesPerPage,
             eventTriggeredDatetime
           )
         else IO.pure(numOfRecentlyUpdatedEntities)
@@ -62,9 +58,9 @@ class Lambda extends LambdaRunner[ScheduledEvent, Int, Config, Dependencies] {
       entitiesClient: EntityClient[IO, Fs2Streams[IO]],
       dADynamoDBClient: DADynamoDBClient[IO],
       dASnsDBClient: DASNSClient[IO],
-      startFrom: Int,
       eventTriggeredDatetime: OffsetDateTime
-  ): IO[Int] =
+  ): IO[Int] = {
+    val ignoredEventTypes = List("Download", "Characterise", "VirusCheck")
     for {
       updatedSinceResponses <- dADynamoDBClient.getItems[GetItemsResponse, PartitionKey](
         List(PartitionKey("LastPolled")),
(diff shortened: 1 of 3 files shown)


scala/lambdas/entity-event-generator-lambda/src/main/scala/uk/gov/nationalarchives/entityeventgenerator/Lambda.scala
```diff
@@ -62,41 +58,46 @@ class Lambda extends LambdaRunner[ScheduledEvent, Int, Config, Dependencies] {
...
+      currentStart = updatedSinceResponse.start
+      recentlyUpdatedEntities <- entitiesClient.entitiesUpdatedSince(updatedSinceAsDate, currentStart)
       _ <- logger.info(s"There were ${recentlyUpdatedEntities.length} entities updated since $updatedSinceAsDate")
 
       entityLastEventActionDate <-
         if (recentlyUpdatedEntities.nonEmpty) {
           val lastUpdatedEntity: Entity = recentlyUpdatedEntities.last
           entitiesClient.entityEventActions(lastUpdatedEntity).map { entityEventActions =>
-            Some(entityEventActions.head.dateOfEvent.toOffsetDateTime)
+            Some(entityEventActions.filterNot(ev => ignoredEventTypes.contains(ev.eventType)).head.dateOfEvent.toOffsetDateTime)
```
sparkhi, line 79: Is it possible that the last entity only has one of the ignored event types? if so, this would not generate correct date.

scala/lambdas/entity-event-generator-lambda/src/main/scala/uk/gov/nationalarchives/entityeventgenerator/Lambda.scala
```diff
@@ -62,41 +58,46 @@ class Lambda extends LambdaRunner[ScheduledEvent, Int, Config, Dependencies] {
...
           _ <- dASnsDBClient.publish[CompactEntity](config.snsArn)(convertToCompactEntities(recentlyUpdatedEntities.toList))
           updateDateAttributeValue = AttributeValue.builder().s(entityLastEventActionDate.get.toString).build()
+          // This is to cover the case where there are more than 1000 items with the same last event action date
+          start = if entityLastEventActionDate.get.isEqual(OffsetDateTime.parse(updatedSinceResponse.datetime)) then currentStart + 1000 else 0
+          startAttributeValue = AttributeValue.builder.n(start.toString).build()
           updateDateRequest = DADynamoDbRequest(
             config.lastEventActionTableName,
             dateItemPrimaryKeyAndValue,
-            Map(datetimeField -> Some(updateDateAttributeValue))
+            Map(datetimeField -> Some(updateDateAttributeValue), "start" -> Some(startAttributeValue))
```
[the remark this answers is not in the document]
MancunianSam: I think those problems were caused by selecting using the global secondary indexes which aren't always populated. I think because this is a single row with two fields and it's rare that we need to increment it anyway, I think it should be ok.

scala/lambdas/entity-event-generator-lambda/src/test/scala/uk/gov/nationalarchives/entityeventgenerator/LambdaSpec.scala
```diff
@@ -15,7 +19,49 @@ class LambdaSpec extends AnyFlatSpec with EitherValues {
...
+    val inputEvent = event("2023-06-07T00:00:00.000000+01:00")
+    val eventActionTime = "2023-06-05T00:00:00.000000+01:00"
+    val dynamoResponse = List(eventActionTime)
+    val entities = List(generateEntity)
+    val (dynamoResult, snsResult, lambdaResult) = runLambda(inputEvent, entities, List(generateEventAction(eventActionTime)), dynamoResponse)
+
+    dynamoResult.head should equal("2023-06-05T00:00+01:00", 1000)
+  }
+
+  "handler" should "write a start number of zero if the original start number was one and the event action time is after the update since time" in {
```
techncl, line 40: It's a bit clearer if you use numbers (0 and 1) here rather than the word form

scala/lambdas/entity-event-generator-lambda/src/main/scala/uk/gov/nationalarchives/entityeventgenerator/Lambda.scala
```diff
@@ -62,41 +58,46 @@ class Lambda extends LambdaRunner[ScheduledEvent, Int, Config, Dependencies] {
...
+            Some(entityEventActions.filterNot(ev => ignoredEventTypes.contains(ev.eventType)).head.dateOfEvent.toOffsetDateTime)
           }
         } else IO.pure(None)
 
       _ <- IO.whenA(entityLastEventActionDate.exists(_.isBefore(eventTriggeredDatetime))) {
         for {
           _ <- dASnsDBClient.publish[CompactEntity](config.snsArn)(convertToCompactEntities(recentlyUpdatedEntities.toList))
           updateDateAttributeValue = AttributeValue.builder().s(entityLastEventActionDate.get.toString).build()
+          // This is to cover the case where there are more than 1000 items with the same last event action date
+          start = if entityLastEventActionDate.get.isEqual(OffsetDateTime.parse(updatedSinceResponse.datetime)) then currentStart + 1000 else 0
```
techncl, line 88: Maybe `currentResultNum`, `currentPosition`,  `nextStart`, `startNext`, `nextStartPoint`, `nextStartingPoint`?

scala/lambdas/entity-event-generator-lambda/src/main/scala/uk/gov/nationalarchives/entityeventgenerator/Lambda.scala
```diff
@@ -62,41 +58,46 @@ class Lambda extends LambdaRunner[ScheduledEvent, Int, Config, Dependencies] {
...
-            Some(entityEventActions.head.dateOfEvent.toOffsetDateTime)
+            Some(entityEventActions.filterNot(ev => ignoredEventTypes.contains(ev.eventType)).head.dateOfEvent.toOffsetDateTime)
           }
         } else IO.pure(None)
 
       _ <- IO.whenA(entityLastEventActionDate.exists(_.isBefore(eventTriggeredDatetime))) {
         for {
           _ <- dASnsDBClient.publish[CompactEntity](config.snsArn)(convertToCompactEntities(recentlyUpdatedEntities.toList))
           updateDateAttributeValue = AttributeValue.builder().s(entityLastEventActionDate.get.toString).build()
+          // This is to cover the case where there are more than 1000 items with the same last event action date
```
techncl, line 87: Should it be "entities" instead of "items"?

techncl (commented):
Looks good! Could we also update the README as it is probably our least descriptive one?

scala/lambdas/entity-event-generator-lambda/src/main/scala/uk/gov/nationalarchives/entityeventgenerator/Lambda.scala
```diff
@@ -62,41 +58,46 @@ class Lambda extends LambdaRunner[ScheduledEvent, Int, Config, Dependencies] {
       entitiesClient: EntityClient[IO, Fs2Streams[IO]],
       dADynamoDBClient: DADynamoDBClient[IO],
       dASnsDBClient: DASNSClient[IO],
-      startFrom: Int,
       eventTriggeredDatetime: OffsetDateTime
-  ): IO[Int] =
+  ): IO[Int] = {
+    val ignoredEventTypes = List("Download", "Characterise", "VirusCheck")
```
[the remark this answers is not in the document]
MancunianSam: 1. Enum is a good idea. I've created one
2. No, it's only this lambda that needs to ignore them, the Preservica client returns the information from Preservica and the calling code decides what to do with it.

scala/lambdas/entity-event-generator-lambda/src/main/scala/uk/gov/nationalarchives/entityeventgenerator/Lambda.scala
MancunianSam: `nextStart` is ok.

PR #564, opened 2025-07-09
Fixes for resender

MancunianSam:
* Update the environment variables to match the state change handler
* Update the config case class to match application.conf
* Use the GSI name from the config
* Create the primary key using the new Dynamo formatter.


techncl (commented):
Looks good! Would you mind updating the env variables in the READMEs for this and the state change handler?

sparkhi (approved):

techncl (approved):
👍

PR #578, opened 2025-07-16
Pin actions to commit hashes

MancunianSam:
Also fix a vulnerability in commons-lang3
Also, as part of trying to fix commons-lang3, I upgraded circe which broke a lot of things. It turns out that circe had nothing to do with the vulnerability but I've changed it all now  to save us a job later.


techncl (approved):
👍

PR #627, opened 2025-09-04
DR2-2429 Entity event generator fixes.

MancunianSam:
The entity event generator fails if the most recent updated item is a
deleted item. This is because there is no information about the entity
type for a deleted item so we can't get the event actions for it.

If there are non-deleted items, we're now filtering out the deleted ones
before we check the event actions.

If there are only deleted items, we set the time to the current time for
Dynamo. This causes the start parameter to increment by 1000 which
allows the entity event generator to progress.

I've changed this so the message it sends is no longer
<entityType>:<ref> for deleted items as we don't know the entity type so
now the id is just the ref.

I've updated the CC queue creator to work with this.

I've also changed the field names from datetime and start because now
we're using the new expressions to update dynamo, we can't use a list of
reserved words and they are both on them.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -13,9 +13,9 @@ jobs:
       next-version: ${{ steps.next-tag.outputs.next-version }}
     steps:
       - uses: actions/checkout@ff7abcd0c3c05ccf6adc123a8cd1fd4fb30fb493
-      - uses: sbt/setup-sbt@3e0b42cddbe318c19f5bbd1af907f913ad005308
+      - uses: sbt/setup-sbt@17575ea4e18dd928fe5968dbe32294b97923d65b
       - name: Configure AWS credentials from management account
-        uses: aws-actions/configure-aws-credentials@7474bc4690e29a8392af63c5b98e7449536d5c3a # v4.3.1
+        uses: aws-actions/configure-aws-credentials@a03048d87541d1d9fcf2ecf528a4a65ba9bd7838 # v5.0.0
         with:
           role-to-assume: arn:aws:iam::${{ secrets.MANAGEMENT_ACCOUNT }}:role/MgmtDPGithubCodeDeploy
           aws-region: eu-west-2
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -31,7 +31,7 @@ jobs:
             print(f"role-name={'${{ github.event.inputs.environment }}'.title()}DPGithubActionsDeployLambdaRole", file=fh)
         shell: python
       - name: Configure AWS credentials for Lambda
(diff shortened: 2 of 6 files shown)


scala/lambdas/entity-event-generator-lambda/src/test/scala/uk/gov/nationalarchives/entityeventgenerator/LambdaSpec.scala
```diff
@@ -27,6 +27,33 @@ class LambdaSpec extends AnyFlatSpec with EitherValues {
     lambdaResult.value should equal(1)
   }
 
+  "handler" should "increment the start argument and send messages if all entities are deleted" in {
```
techncl, line 30: 1. To match the test below, maybe it could be "increment the start argument, not update the event datetime and send messages..."
2. Maybe "startAt argument" instead of "start argument"?

scala/lambdas/custodial-copy-queue-creator/src/main/scala/uk/gov/nationalarchives/custodialcopyqueuecreator/Lambda.scala
```diff
@@ -51,30 +50,36 @@ object Lambda:
...
     }
 
   enum MessageBody(val id: UUID, val deleted: Boolean):
-    case IoMessageBody(override val id: UUID, override val deleted: Boolean) extends MessageBody(id, deleted)
-    case CoMessageBody(override val id: UUID, override val deleted: Boolean) extends MessageBody(id, deleted)
-    case SoMessageBody(override val id: UUID, override val deleted: Boolean) extends MessageBody(id, deleted)
+    case IoMessageBody(override val id: UUID) extends MessageBody(id, false)
+    case CoMessageBody(override val id: UUID) extends MessageBody(id, false)
+    case SoMessageBody(override val id: UUID) extends MessageBody(id, false)
+    case DeletedMessageBody(override val id: UUID) extends MessageBody(id, true)
```
techncl, line 85: Maybe `DeletedEntityMessageBody`?

techncl (commented):
Looks good! Just some minor comments. Also,

1. Should we log any of the deletions?
2. Would you mind updating the README?

scala/lambdas/custodial-copy-queue-creator/src/main/scala/uk/gov/nationalarchives/custodialcopyqueuecreator/Lambda.scala
MancunianSam: It's a bit long. I'll change it to `DeletionMessageBody` to match the change on the CC PR.

techncl (approved):
👍

techncl (commented):
Just a few minor comments

techncl (approved):
👍

PR #635, opened 2025-09-09
DR2-2434 Fix E2E tests

MancunianSam:
This allows them to work with the new multiple files structure.

The metadata is now an array and the files is stored in
`<assetId>/<fileId>` in S3.

I've increased the timeout as sometimes we miss the first pass of
postingest and have to wait an hour.


techncl (commented):
Looks good! Just 2 minor comments.

PR #651, opened 2025-09-16
Entity event generator fixes

MancunianSam:
There are two changes here.

We're now returning `hasNext:Boolean` which shows whether there is
another page. If there is, or all messages are deleted, we keep the next
updated time the same and increment the page.

The other change is to return empty list if the last updated date is
after the event time. This is to stop the lambda recursively trying
again until the event date is later as this is a bit of a waste. Now the
lambda will complete and we can try again on the next invocation.

### DIFF
--- a/scala/lambdas/entity-event-generator-lambda/src/main/scala/uk/gov/nationalarchives/entityeventgenerator/Lambda.scala
+++ b/scala/lambdas/entity-event-generator-lambda/src/main/scala/uk/gov/nationalarchives/entityeventgenerator/Lambda.scala
@@ -65,7 +65,7 @@ class Lambda extends LambdaRunner[ScheduledEvent, Int, Config, Dependencies] {
   ): IO[Int] = {
     val ignoredEventTypes = IgnoredEventTypes.values.map(_.toString)
     for {
-      _ <- logger.info(s"Getting entities for triggered time $eventTriggeredDatetime")
+      _ <- logger.info(s"Getting entities - event triggered time: $eventTriggeredDatetime")
       updatedSinceResponses <- dADynamoDBClient.getItems[GetItemsResponse, PartitionKey](
         List(PartitionKey("LastPolled")),
         config.lastEventActionTableName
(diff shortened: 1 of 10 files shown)


scala/lambdas/entity-event-generator-lambda/src/main/scala/uk/gov/nationalarchives/entityeventgenerator/Lambda.scala
```diff
@@ -65,45 +65,50 @@ class Lambda extends LambdaRunner[ScheduledEvent, Int, Config, Dependencies] {
   ): IO[Int] = {
     val ignoredEventTypes = IgnoredEventTypes.values.map(_.toString)
     for {
+      _ <- logger.info(s"Getting entities for triggered time $eventTriggeredDatetime")
```
techncl, line 68: 1. Maybe `s"Getting entities for event triggered time $eventTriggeredDatetime"`?
2. I'm not sure I understand this because it reads like we're getting entities with an updatedSince of `eventTriggeredDatetime`; should this message read something like `s"Getting entities at event triggered time $eventTriggeredDatetime"`, `s"Event triggered time: $eventTriggeredDatetime - getting entities"` or  `s"Getting entities - event triggered time: $eventTriggeredDatetime"`

scala/lambdas/entity-event-generator-lambda/src/main/scala/uk/gov/nationalarchives/entityeventgenerator/Lambda.scala
```diff
@@ -65,45 +65,50 @@ class Lambda extends LambdaRunner[ScheduledEvent, Int, Config, Dependencies] {
...
       updatedSinceResponses <- dADynamoDBClient.getItems[GetItemsResponse, PartitionKey](
         List(PartitionKey("LastPolled")),
         config.lastEventActionTableName
       )
       updatedSinceResponse = updatedSinceResponses.head
```
techncl, line 73: Since we only want 1 result, should we add a limit on this `getItems` query that will only get one item?

scala/lambdas/entity-event-generator-lambda/src/main/scala/uk/gov/nationalarchives/entityeventgenerator/Lambda.scala
```diff
@@ -65,45 +65,50 @@ class Lambda extends LambdaRunner[ScheduledEvent, Int, Config, Dependencies] {
...
             Map(datetimeField -> updateDateAttributeValue, "startAt" -> startAttributeValue)
           )
-          dynamoStatusCode <- dADynamoDBClient.updateAttributeValues(updateDateRequest)
-          _ <- logger.info(s"Dynamo updateAttributeValues returned status code $dynamoStatusCode")
-        } yield ()
-      }
-    } yield recentlyUpdatedEntities.length
+          for
+            _ <- dASnsDBClient.publish[CompactEntity](config.snsArn)(convertToCompactEntities(recentlyUpdatedEntities.toList))
+            dynamoStatusCode <- dADynamoDBClient.updateAttributeValues(updateDateRequest)
```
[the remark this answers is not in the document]
MancunianSam: We do because the startAt field needs to be updated if that happens. I suppose we could remove the `datetimeField` update but it seems needlessly complicated.

scala/lambdas/entity-event-generator-lambda/src/main/scala/uk/gov/nationalarchives/entityeventgenerator/Lambda.scala
MancunianSam: I wasn't thinking that hard about it. I wanted the event triggered time in there. I've changed it to `s"Getting entities - event triggered time: $eventTriggeredDatetime"`

scala/lambdas/entity-event-generator-lambda/src/main/scala/uk/gov/nationalarchives/entityeventgenerator/Lambda.scala
MancunianSam: Doesn't seem worth it really. We know there's only one row in there.

techncl (approved):
👍

PR #654, opened 2025-09-25
Update diagram and exclude docs directory

techncl:


MancunianSam (commented):
Could you add the DRI preingest and the CC reconciler to the diagram please.

PR #655, opened 2025-09-25
Update package builder code and docs

techncl:


MancunianSam (approved):
We can merge this one but we're spending far too much time renaming variables and method names so lets make this the last one for a while..

PR #1004, opened 2026-04-13
Allow e2e tests on staging.

MancunianSam:
This creates the e2e tests role for staging and adds it to the KMS key.

Also, if max concurrency is set to one, it can cause problems where the
next step function run isn't triggered so I've moved the numbers around
a bit.

### DIFF
--- a/terraform/common.tf
+++ b/terraform/common.tf
@@ -7,7 +7,7 @@ locals {
   ingest_run_workflow_step_function_name               = "${local.environment}-dr2-ingest-run-workflow"
   additional_user_roles                                = local.environment != "prod" ? [data.aws_ssm_parameter.dev_admin_role.value] : []
   anonymiser_roles                                     = local.environment == "intg" ? flatten([module.dr2_court_document_package_anonymiser_lambda.*.lambda_role_arn]) : []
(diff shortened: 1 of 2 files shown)


terraform/common.tf
```diff
@@ -7,7 +7,7 @@ locals {
   ingest_run_workflow_step_function_name               = "${local.environment}-dr2-ingest-run-workflow"
   additional_user_roles                                = local.environment != "prod" ? [data.aws_ssm_parameter.dev_admin_role.value] : []
   anonymiser_roles                                     = local.environment == "intg" ? flatten([module.dr2_court_document_package_anonymiser_lambda.*.lambda_role_arn]) : []
-  e2e_test_roles                                       = local.environment == "intg" ? [module.dr2_run_e2e_tests_role[0].role_arn] : []
+  e2e_test_roles                                       = local.environment == "prod" ? [] : [module.dr2_run_e2e_tests_role[0].role_arn]
```
techncl, line 10: Would this also allow for e2e test roles on other environments like sandbox? If so, would that be an issue?

techncl (commented):

terraform/common.tf
MancunianSam: There are no roles for sandbox that can deploy the ingest code so unless we did something very silly like deploying to sandbox with the admin credentials then I think this is ok.

techncl (approved):
👍

PR #1016, opened 2026-04-16
DDSM-400 roles anywhere add forgotten variables for workflow secrets

James-Noble:
should fix the environment secret issue I was having

btw interesting thing, terraform plan failed but the cat exit-code thing after it didn't recognise it as a failure

### DIFF
--- a/terraform/custodial_copy.tf
+++ b/terraform/custodial_copy.tf
@@ -3,6 +3,19 @@ locals {
   custodial_copy_db_builder_queue_name = "${local.custodial_copy_name}-db-builder"
 }
 
+variable "site_outbound_subnet" {
+  type        = string
+  sensitive   = true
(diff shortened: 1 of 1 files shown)


terraform/custodial_copy.tf
```diff
@@ -3,6 +3,19 @@ locals {
...
+  description = "IP address that web access from the Kew site goes through - provided via Github Secrets"
+  default     = ""
+}
+
+variable "custodial_copy_x509_subject_cn" {
+  type        = string
+  sensitive   = true
+  description = "x509 Subject CN of the certificate user will auth with to Roles Anywhere - provided via Github Secrets"
+}
+
```
MancunianSam, line 18: Could you create a new file called `root_variables.tf` and put these two in there please. I'm doing something else that needs a variable and it's nice to keep them in one place.

James-Noble:
superseded by pull request #1017

PR #1017, opened 2026-04-17
DDSM-400 Roles anywhere with ssm params instead of github secrets

James-Noble:
Based on advice from Sam I'm using ssm params instead of github secrets for the certificate common name and the ip locking subnet, since otherwise it becomes difficult to run terraform plan etc. on intg environment without being prompted for the variables

Now that I know how, I've tested this with plan + apply on the intg environment, then a successive plan which is clean.  I've tested acquiring the credential from the DRI server and it worked :)

Also fixed issue where policy template couldn't accept list of string (I needed to jsonencode it and remove quote marks in the policy template), and upgraded aws provider to .41 since it wouldn't let me proceed without that

Will close the previous pull request as that's only to add stuff relating to variables and github secrets, which are now rid of


MancunianSam (approved):
👍

PR #1018, opened 2026-04-17
DR2-2784 Ingest to use lambda aliases

MancunianSam:
* All lambdas now deploy from S3 in terraform
* All SFN lamdbas publish new versions on code change
* When a new version is published, an alias is created based on the
  terraform variable.
* The deploy step which updates the function code has been deleted.
* I've added a lambda which will periodically delete any versions where
  there are more than 3 versions and they are more than a month old.


terraform/lambda_version_delete.tf
```diff
@@ -0,0 +1,32 @@
...
+  source                  = "git::https://github.com/nationalarchives/da-terraform-modules//cloudwatch_events?ref=main"
+  rule_name               = "${local.delete_lambda_version_name}-schedule"
+  schedule                = "rate(1 day)"
+  lambda_event_target_arn = "arn:aws:lambda:eu-west-2:${data.aws_caller_identity.current.account_id}:function:${local.entity_event_lambda_name}"
+}
+
+data "archive_file" "delete_lambda_version_code" {
+  type        = "zip"
+  source_file = "${path.module}/templates/lambda/delete_lambda_versions.py"
+  output_path = "${path.module}/lambda/function.zip"
```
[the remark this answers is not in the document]
MancunianSam: I've changed it to write to the templates/lambda directory and added this to .gitignore.

terraform/preingest/preingest.tf
```diff
@@ -72,18 +76,29 @@ module "dr2_preingest_aggregator_lambda" {
...
 }
 
+resource "terraform_data" "create_lambda_alias" {
+  triggers_replace = [
+    module.dr2_preingest_package_builder_lambda.lambda_function.version
+  ]
+  provisioner "local-exec" {
+    command = "aws lambda create-alias --function-name ${local.package_builder_lambda_name} --name ${local.alias_name} --function-version ${module.dr2_preingest_package_builder_lambda.lambda_function.version} || true"
+  }
+}
```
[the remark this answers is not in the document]
MancunianSam: I've changed this to try update and then add if that fails as suggested
```bash
aws lambda update-alias --function-name ${each.key} --name ${local.lambda_alias_name} --function-version ${each.value} || \
aws lambda create-alias --function-name ${each.key} --name ${local.lambda_alias_name} --function-version ${each.value}
```

terraform/templates/lambda/delete_lambda_versions.py
```diff
@@ -0,0 +1,19 @@
+import boto3
+from datetime import datetime, timezone, timedelta
+
+client = boto3.client("lambda")
+
+def lambda_handler(event, context):
+    functions = client.list_functions()['Functions']
+    for function in functions:
+        function_name = function['FunctionName']
+        versions = client.list_versions_by_function(FunctionName=function_name)['Versions']
```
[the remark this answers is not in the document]
MancunianSam: I've added pagination for list_functions and list_versions_by_function

terraform/templates/lambda/delete_lambda_versions.py
```diff
@@ -0,0 +1,19 @@
...
+        named_versions = [v for v in versions if v['Version'] != '$LATEST']
+        named_versions_length = len(named_versions)
+        if named_versions_length > 3:
+            to_delete = named_versions[:(named_versions_length - 3)]
+            for version_to_delete in to_delete:
+                lmd = datetime.strptime(version_to_delete['LastModified'], "%Y-%m-%dT%H:%M:%S.%f%z")
+                cutoff = datetime.now(timezone.utc) - timedelta(days=30)
+                if lmd < cutoff:
+                    client.delete_function(FunctionName=function_name, Qualifier=version_to_delete['Version'])
```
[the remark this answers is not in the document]
MancunianSam: I've added a tag into the terraform called SfnFunction and then in this script, I filter for that tag. I've also added error handling to the delete call so a single failed delete won't prevent others being tried.

.github/workflows/deploy.yml
```diff
@@ -88,33 +98,24 @@ jobs:
...
-          role-session-name: ECRLogin
-      - run: | 
-          pip install boto3
-          git tag $TO_DEPLOY
-          git push origin $TO_DEPLOY
-          python .github/scripts/deploy.py $ENVIRONMENT $TO_DEPLOY
+      - run: python .github/scripts/send_message.py
+        env:
+          MESSAGE: ":white_check_mark: ingest deploy successful for ${{ github.event.inputs.environment }} for version ${{ github.event.inputs.to-deploy }}"
+      - run: gh workflow run e2e-tests.yml -f environment=$ENVIRONMENT -f to-deploy=$TO_DEPLOY --ref DR2-2784-ingest-to-use-lambda-aliases
```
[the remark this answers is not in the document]
MancunianSam: I didn't mean to leave this here.

terraform/lambda_version_delete.tf
```diff
@@ -0,0 +1,32 @@
+locals {
+  delete_lambda_version_name = "${local.environment}-dr2-delete-lambda-version"
+}
+module "dr2_delete_lambda_version_schedule" {
+  source                  = "git::https://github.com/nationalarchives/da-terraform-modules//cloudwatch_events?ref=main"
+  rule_name               = "${local.delete_lambda_version_name}-schedule"
+  schedule                = "rate(1 day)"
+  lambda_event_target_arn = "arn:aws:lambda:eu-west-2:${data.aws_caller_identity.current.account_id}:function:${local.entity_event_lambda_name}"
```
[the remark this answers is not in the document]
MancunianSam: This was a copy and paste error.

terraform/common.tf
```diff
@@ -460,6 +460,27 @@ module "sample_files_bucket" {
...
+    module.dr2_ingest_parent_folder_opex_creator_lambda.lambda_function,
+    module.dr2_ingest_asset_reconciler_lambda.lambda_function,
+    module.dr2_ingest_flow_control_lambda.lambda_function,
+  ] : l.function_name => l }
+  triggers_replace = [
+    each.value.version
+  ]
+
+  provisioner "local-exec" {
+    command = "aws lambda create-alias --function-name ${each.key} --name ${local.lambda_alias_name} --function-version ${each.value.version} || true"
```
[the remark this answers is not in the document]
MancunianSam: Fixed as per above comment.

terraform/templates/iam_policy/version_delete.json.tpl
```diff
@@ -0,0 +1,28 @@
...
+  "Statement": [
+    {
+      "Action": [
+        "lambda:ListFunctions",
+        "lambda:ListVersionsByFunction",
+        "lambda:DeleteFunction"
+      ],
+      "Effect": "Allow",
+      "Resource": ["arn:aws:lambda:eu-west-2:${account_id}:function:*", "arn:aws:lambda:eu-west-2:${account_id}:function:*:*"],
+      "Sid": "updateLambdaVersions"
```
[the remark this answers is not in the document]
MancunianSam: I've restricted this to a list of lambda arns which will need the versions cleaning up and moved ListFunctions to its own statement.

terraform/root_variables.tf
```diff
@@ -0,0 +1 @@
+variable "lambda_code_version" {}
```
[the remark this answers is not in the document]
MancunianSam: I can live with this being mandatory. I've added a `TF_VAR_lambda_code_version` env var in the check terraform drift workflow.

.github/workflows/check-terraform-drift.yml
```diff
@@ -27,6 +27,7 @@ jobs:
         run: |
           terraform init
           terraform workspace select prod
+          export TF_VAR_lambda_code_version=$(curl https://api.github.com/repos/nationalarchives/dr2-ingest/tags | jq -r '.[0].name')
```
kurtismash: Potential to it the rate limit as this is an anomalous request, best to use `gh` cli with the workflow token?

.github/workflows/test_plan.yml
```diff
@@ -28,5 +25,6 @@ jobs:
       - run: |
           terraform init
           terraform workspace select intg
-          terraform plan
+          export TF_VAR_lambda_code_version=$(curl https://api.github.com/repos/nationalarchives/dr2-ingest/tags | jq -r '.[0].name')
```
kurtismash: Potential to it the rate limit as this is an anomalous request, best to use gh cli with the workflow token?

terraform/postingest/postingest.tf
```diff
@@ -14,6 +16,7 @@ locals {
   six_hours                  = 60 * 60 * 6
   seven_days                 = 60 * 60 * 24 * 7
   messages_visible_threshold = 1000000
+  code_deploy_bucket         = "mgmt-dp-code-deploy"
```
kurtismash: Should this be a variable to the module?

terraform/preingest/preingest.tf
```diff
@@ -18,6 +18,8 @@ locals {
   aggregator_group_size                        = 10000                                                                                      # Max size of an aggregation group.
   aggregator_queue_visibility_timeout          = local.aggregator_primary_grouping_window_seconds + local.aggregator_lambda_timeout_seconds # <=43200 for SQS.
   messages_visible_threshold                   = 1000000
+  code_deploy_bucket                           = "mgmt-dp-code-deploy"
```
kurtismash, line 21: Should this be a module variable?

terraform/preingest/preingest.tf
```diff
@@ -73,18 +77,32 @@ module "dr2_preingest_aggregator_lambda" {
...
+  triggers_replace = [
+    module.dr2_preingest_package_builder_lambda.lambda_function.version
+  ]
+  provisioner "local-exec" {
+    command = <<-EOT
+      aws lambda update-alias --function-name ${local.package_builder_lambda_name} --name ${local.alias_name} --function-version ${module.dr2_preingest_package_builder_lambda.lambda_function.version} || \
+      aws lambda create-alias --function-name ${local.package_builder_lambda_name} --name ${local.alias_name} --function-version ${module.dr2_preingest_package_builder_lambda.lambda_function.version}
+    EOT
+  }
+}
```
kurtismash: Is it worth wrapping this in a module to avoid duplicating all the bash commands?

terraform/preingest/preingest.tf
```diff
@@ -73,18 +77,32 @@ module "dr2_preingest_aggregator_lambda" {
...
     account_id                  = data.aws_caller_identity.current.account_id
     package_builder_lambda_name = local.package_builder_lambda_name
     retry_statement             = jsonencode([{ ErrorEquals = ["States.ALL"], IntervalSeconds = 2, MaxAttempts = 6, BackoffRate = 2, JitterStrategy = "FULL" }])
+    alias_name                  = local.alias_name
   })
   step_function_name = local.preingest_name
   step_function_role_policy_attachments = {
     preingest_step_function_policy = module.dr2_preingest_step_function_policy.policy_arn
   }
+  depends_on = [terraform_data.create_lambda_alias]
```
kurtismash: I think this will cause the module to be updated each time the alias is changed.

.github/workflows/e2e-tests.yml
```diff
@@ -26,15 +33,25 @@ jobs:
...
-          JUDGMENT_SQS_QUEUE: https://sqs.eu-west-2.amazonaws.com/${{ secrets.ACCOUNT_NUMBER }}/intg-dr2-preingest-courtdoc-importer
+          INGEST_SQS_QUEUE: https://sqs.eu-west-2.amazonaws.com/${{ secrets.ACCOUNT_NUMBER }}/${{ github.event.inputs.environment }}-dr2-preingest-tdr-importer
+          EXTERNAL_LOG_GROUP: arn:aws:logs:eu-west-2:${{ secrets.ACCOUNT_NUMBER }}:log-group:/${{ github.event.inputs.environment }}-external-notifications
+          PREINGEST_SFN_ARN: arn:aws:states:eu-west-2:${{ secrets.ACCOUNT_NUMBER }}:stateMachine:${{ github.event.inputs.environment }}-dr2-preingest-tdr
+          COPY_FILES_LOG_GROUP: arn:aws:logs:eu-west-2:${{ secrets.ACCOUNT_NUMBER }}:log-group:/aws/lambda/${{ github.event.inputs.environment }}-dr2-preingest-tdr-importer
+          JUDGMENT_SQS_QUEUE: https://sqs.eu-west-2.amazonaws.com/${{ secrets.ACCOUNT_NUMBER }}/${{ github.event.inputs.environment }}-dr2-preingest-courtdoc-importer
+          S3_BUCKET: ${{ github.event.inputs.environment }}-dr2-ingest-raw-cache
+          LOCK_TABLE: ${{ github.event.inputs.environment }}-dr2-ingest-lock
+      - run: gh workflow run deploy.yml -f environment=staging -f to-deploy=$TO_DEPLOY
```
techncl, line 50: Is it ok that the environment if fixed to staging?

.github/workflows/build.yml
```diff
@@ -49,7 +49,7 @@ jobs:
           zip -jA court-document-package-anonymiser rust/lambdas/courtdoc-anonymiser/target/lambda/lambda/bootstrap
           cp court-document-package-anonymiser target/outputs
           cd target/outputs
```
techncl: Should "target/outputs" be passed into the python script (instead of hardcoding on line 8)?

.github/scripts/upload_files.py
```diff
@@ -0,0 +1,24 @@
...
+files = os.listdir(path)
+client = boto3.client("s3")
+version = sys.argv[1]
+
+def upload_file(file):
+    with open(f"{path}/{file}", "rb") as f:
+        digest = hashlib.sha256(f.read()).digest()
+        sha256 = base64.b64encode(digest).decode()
+        client.put_object(Bucket="mgmt-dp-code-deploy", Key=f"{version}/{file}", Body=f, ChecksumSHA256=sha256)
+        print(f"Uploaded {file}")
```
techncl: Is it worth appending the bucket to this message?

techncl (commented):
Looks good! Just some minor comments.

terraform/postingest/postingest.tf
MancunianSam: It never changes but I suppose it's cleaner as a variable. I've added one.

terraform/preingest/preingest.tf
MancunianSam: Sure, why not.

terraform/preingest/preingest.tf
MancunianSam: Isn't that what we want though, we want the lambdas in the step function to be updated when the alias changes?

.github/workflows/e2e-tests.yml
MancunianSam: Yep, it only runs when the environment is intg. I could probably have a map somewhere of {intg -> staging, staging -> prod} but this is cleaner.

.github/workflows/build.yml
MancunianSam: Yeah good point. It wouldn't actually work as I've got it here either. I've added it as an argument.

kurtismash (approved):
I'm still not sure about the `depends_on = [module...]` bit on the Step Functions but we can fix that later if it becomes an issue. Depending on a module is not recommended.

> You should only use depends_on as a last resort because it can cause Terraform to create more conservative plans that replace more resources than necessary. For example, Terraform may treat more values as unknown "(known after apply)" because it is uncertain what changes will occur on the upstream object. This is especially likely when you use depends_on for modules. ([link](https://developer.hashicorp.com/terraform/language/meta-arguments/depends_on#processing-and-planning-consequences))

If changes are needed, I'd advise passing `local.alias` into the `terraform_data` module+resource, then using it from the `terraform_data` outputs.

PR #1019, opened 2026-04-20
DR2-2263 Send IngestStarted message.

MancunianSam:
This will send an IngestStarted update message directly after the item
is written to the lock table. This means that it's possible for us to
send an IngestStarted message and then the step function invocation
fails. This is fine.

I've changed the tests slightly so the helper method returns a case
class.


techncl (approved):
👍

PR #1023, opened 2026-04-22
DR2-2786 Add wiz org role to records metadata key.

MancunianSam:
Wiz can't scan this bucket as its role hasn't been added to the KMS key.


techncl (approved):
👍

PR #1026, opened 2026-04-22
Remove parliament roles from raw cache policy.

MancunianSam:
Wiz was complaining because these roles haven't been used for half a
year so I removed them but now the bucket policy won't apply because
they don't exist.

There's no reason to have these here for now so I've removed them.

This will defautlt to the standard DenyNonSsl policy now.


techncl (approved):
👍

PR #1027, opened 2026-04-23
Use custom encoders

MancunianSam:
We were using the generic encoders which encode the enums as objects
instead of as strings. This is now using the correct encoder.


techncl (approved):
👍

PR #1028, opened 2026-04-23
Tag with environment name

MancunianSam:
We're tagging with the version in the build workflow. We need to tag
with the environment name as well so we can see which environments are
where.


techncl (approved):
👍

PR #1029, opened 2026-04-24
DR2-2555 Move alerts to general channel.

MancunianSam:
* All alarm alerts
* Step function failures
* Guard duty alerts
* IP lock checker failures
* Ingest paused notifications


techncl (approved):
👍

PR #1030, opened 2026-04-24
Add lifecyle rule to delete delete object markers.

MancunianSam:
There are ~400000 delete object markers in this bucket which it would be
nice to get rid of.


techncl (approved):
👍

PR #1051, opened 2026-05-01
Give permissions to e2e role to adhoc infra in intg

sparkhi:


terraform/dri_migration.tf
```diff
@@ -12,6 +12,7 @@ module "dr2_dri_migration_key" {
     user_roles = [
       module.dr2_dri_migration_role.role_arn,
       module.dri_preingest.importer_lambda.role,
+      local.e2e_test_roles[0]
```
MancunianSam: This will fail on prod because `local.e2e_test_roles` is empty. You need to set it like it is for the main kms key, something like:
```hcl
concat([
      module.dr2_dri_migration_role.role_arn,
      module.dri_preingest.importer_lambda.role
    ], local.e2e_test_roles)
```

terraform/ad_hoc_ingest.tf
```diff
@@ -9,6 +9,7 @@ module "dr2_archivists_key" {
     user_roles = [
       data.aws_ssm_parameter.archivist_role.value,
       module.ad_hoc_preingest.importer_lambda.role,
+      local.e2e_test_roles[0]
```
MancunianSam: Same here, if you change it to use concat

terraform/templates/iam_policy/e2e_tests_policy.json.tpl
```diff
@@ -37,7 +37,9 @@
       ],
       "Resource": [
         "${external_notifications_log_group}",
-        "${copy_files_from_tdr_log_group}"
+        "${copy_files_from_tdr_log_group}",
```
MancunianSam, line 40: We've added monitoring for the courtdoc log group in here now so can you add that group here.

MancunianSam (approved):

PR #1056, opened 2026-05-14
DR2-2834 Filter out mock series.

MancunianSam:
I've tested this and it seems to work.


terraform/templates/sns/tdr_preingest_filter_policy.json.tpl
kurtismash, line 1: Should this file have a more meaningful name? It looks like it's the SNS filer policy for TDR preingest.

terraform/templates/sns/tdr_preingest_filter_policy.json.tpl
```diff
@@ -1,3 +1,8 @@
 {
-  "bucket": ["${bucket}"]
+  "bucket": ["${bucket}"],
```
kurtismash, line 2: Why are we filtering on bucket to begin with? Not the event type or something more descriptive?

MancunianSam: It was a check against us listening for messages from the wrong environment I think. This is what TDR send us:
```json
{
  "bucket": "tdr-export-staging",
  "assetId": "b04745c4-9e0b-4221-b2da-3dfa442abffc",
  "fileId": "182a46e4-8350-43eb-8c90-e8585c45e7df",
  "transferringBody": "Mock 1 Department",
  "series": "TEST 123"
}
```
There's not much else in there to filter on.

terraform/templates/sns/tdr_preingest_filter_policy.json.tpl
MancunianSam: Yeah ok.

terraform/templates/sns/tdr_preingest_filter_policy.json.tpl
kurtismash: We'll need to change this before TDR start sending other messages...

kurtismash (approved):

PR #1084, opened 2026-06-01
Dr2 2375 rename postingest ddb gsi name

techncl:


scala/lambdas/postingest-message-resender/src/main/resources/application.conf
```diff
@@ -1,3 +1,3 @@
 state-table-name = ${POSTINGEST_STATE_DDB_TABLE}
-state-gsi-name = ${POSTINGEST_DDB_TABLE_BATCHPARENT_GSI_NAME}
+state-gsi-name = ${POSTINGEST_DDB_TABLE_QUEUED_AT_GSI_NAME}
```
sparkhi: I know the ticket mentioned this name, but should it be more specific with LAST_QUEUED because there is  `firstQueued` in the table too.

sparkhi (commented):
looks good, just one minor afterthought.

scala/lambdas/postingest-message-resender/src/main/resources/application.conf
techncl: This is what I thought too. I'll message Kurtis just in case there was a specific reason why he wanted it to be generic

sparkhi (approved):

PR #1085, opened 2026-06-02
Dr2 2582 rename executions running by source system metrics

techncl:


python/lambdas/ingest-metric-collector/ingest_metric_collector_test.py
```diff
@@ -65,26 +89,11 @@ def test_get_stepfunction_metrics_should_return_metrics_when_executions_and_sour
...
         # 1 metric for total executions + metrics per source system
-        self.assertEqual(1 + len(ingest_metric_collector.SOURCE_SYSTEMS), len(metrics))  # total + sources + DEFAULT
-
-        # Check that TDR, COURTDOC and DEFAULT, all got Value as 1
-        tdr_metric = next(
-            m for m in metrics
-            if {"Name": "SourceSystem", "Value": "TDR"} in m["Dimensions"]
-        )
-        self.assertEqual(1, tdr_metric["Value"])
+        self.assertEqual(1 + 5, len(metrics))  # total + sources + DEFAULT
```
sparkhi: - maybe use length of `expected_source_systems` instead of hardcoded number 5

- The inline comment `# total + sources + DEFAULT` can go away as the comment above is fine and `DEFAULT` is included in the list as a valid source system

python/lambdas/ingest-metric-collector/ingest_metric_collector_test.py
```diff
@@ -4,8 +4,25 @@
...
+        "Unit": unit
+    }
+    if source_system:
+        metrics["Dimensions"].append({"Name": "SourceSystem", "Value": source_system})
+
+    return metrics
+
 
 class TestLambdaFunction(unittest.TestCase):
+    expected_source_systems = ("TDR", "COURTDOC", "ADHOC", "PA", "DEFAULT")
```
sparkhi, line 25: We now have this same list in test class against implementation, should we just use the one from implementation as that list governs our expectations. (will also relate to a comment further down)

sparkhi (commented):
Just a couple of minor comments

python/lambdas/ingest-metric-collector/ingest_metric_collector_test.py
techncl: Do you mean using `SOURCE_SYSTEMS` instead? I didn't want to do this because I want, if we are to update the source systems in the future, for it to be a conscious decision. If we relied on `SOURCE_SYSTEMS` as our source of truth, if one of the source systems gets modified/removed by accident (and we missed it), the tests would still pass but I want it to fail and have us alerted to it; this is a similar reason as to why I refactored these tests as we weren't testing that the structure of the metrics is how we expect it should be. It does add redundancy but at least we can be sure that the change was intentional.

sparkhi: That's fine

sparkhi (approved):

PR #1087, opened 2026-06-05
Add Intelligent Caching ADR

techncl:


docs/architecture-decision-records/0027-intelligent-caching.md
```diff
@@ -0,0 +1,48 @@
+# 27. Custodial Copy Intelligent Caching
+
+**Date:** 2026-06-04
+
+## Context
+As we [migrate from DRI](/python/dri-migration/README.md), we will be uploading files to Preservica (DR2) from Kew.
```
sparkhi: I think, we use the term "Preservation System" instead of "Preservica" to mean our ADRs are not dependent on the underlying system. It will need to change in multiple places within this doc.

docs/architecture-decision-records/0027-intelligent-caching.md
```diff
@@ -0,0 +1,48 @@
+# 27. Custodial Copy Intelligent Caching
+
+**Date:** 2026-06-04
+
+## Context
+As we [migrate from DRI](/python/dri-migration/README.md), we will be uploading files to Preservica (DR2) from Kew. 
+Custodial Copy (CC), in its current state, downloads the file from Preservica and store it in Kew which, for the files we
```
sparkhi: should we say `file(s)` 
stores - typo missing s

docs/architecture-decision-records/0027-intelligent-caching.md
```diff
@@ -0,0 +1,48 @@
...
+want to migrate from DRI, wastes resources, time, and adds extra cost and time since the files are already in Kew.
```
sparkhi: adds extra cost ~and time~ you have already covered time before

docs/architecture-decision-records/0027-intelligent-caching.md
```diff
@@ -0,0 +1,48 @@
...
+
+## Decision
+
+We would want to establish the following arrangement: files that need to end up being stored in Kew will be held in a storage
+location (folder) in Kew. They might be held in arbitrary folder structures within this location.
+
+When CC is about to download a file from Preservica, it will check this local cache to see if the exact file
```
sparkhi: Its probably useful to say how it is checking exact file i.e. `asset_id` + `file_id` rather than something like checksum. I assume, if a same file appears in more than one place then, there will be those many entries in the cache,

docs/architecture-decision-records/0027-intelligent-caching.md
```diff
@@ -0,0 +1,48 @@
...
+
+| file_id (type text)                  | file_path (type text) | asset_id (type text)                 |
+|--------------------------------------|-----------------------|--------------------------------------|
+| 738c55cf-cfa1-4f57-98c9-c585577b9916 | path/to/local/file1   | 24fce28a-4605-4071-922b-f70ab12bcbe4 |
+| f07e6bb6-74b1-4607-af70-93bb6045d716 | path/to/local/file2   | 24fce28a-4605-4071-922b-f70ab12bcbe4 |
+
+
+### Implementation
+
+- In the DRI migration script, after the writing of the results to JSON files in S3, we will add code that writes the
```
sparkhi: I think, in addition to DRI migration, this could be used by ad hoc ingest process too, unless we decide to leave it out. Until now, ad-hoc ingests have dealt with small data but I imagine, ingesting something like a big video file will benefit from this process

sparkhi (commented):
Looks good to me, some minor questions and suggestions

docs/architecture-decision-records/0027-intelligent-caching.md
techncl: I'll just change it to "files" and modify the sentence to make sense

docs/architecture-decision-records/0027-intelligent-caching.md
techncl: Should I add something like "In the DRI migration script (and potentially other scripts in the future, for e.g. adhoc ingests)"?

sparkhi: That would be great. That wording tells that the decision is made to keep it extendable architecture 👍🏼

docs/architecture-decision-records/0027-intelligent-caching.md
```diff
@@ -0,0 +1,53 @@
...
+| 738c55cf-cfa1-4f57-98c9-c585577b9916 | path/to/local/file1   | 24fce28a-4605-4071-922b-f70ab12bcbe4 |
+| f07e6bb6-74b1-4607-af70-93bb6045d716 | path/to/local/file2   | 24fce28a-4605-4071-922b-f70ab12bcbe4 |
+
+We would also want to begin the upload and ingest process from this cache location.
+
+### Implementation
+
+- In the DRI migration script (and potentially other scripts in the future, for e.g. [Adhoc Ingest](https://github.com/nationalarchives/dr2-ingest/tree/main/python/ad-hoc-ingest)), 
+  after the writing of the results to JSON files in S3, we will add code that writes the (local) file path, file id and 
+  asset id of each asset to a SQLite database (the path of which would be passed in when the script is run).
```
kurtismash, line 42: Is this the correct order? Are we happy to accept that CC may download items if they ingest into the preservation system before we've finished the upload/migration?

docs/architecture-decision-records/0027-intelligent-caching.md
```diff
@@ -0,0 +1,53 @@
...
+When CC is about to download a file from the Preservation System, it will check this local cache to see if the exact file 
+is already available there, and sources it locally if so, otherwise, it will source it from the Preservation System.
+It has been decided that sourcing the files will be done via fileId rather than a checksum due to: simplicity as files can
+have multiple checksums so searching becomes more complex, the files already have a file id readily available in 
+both the migration script and in Custodial Copy, using a file id allows for faster database searching as the file id column
+would be unique whereas a checksum column couldn't be as you can have multiple files with the same checksum and using the 
+file's id requires fewer code changes than using a checksum.
+
+Due to the simplicity of this approach, the size of the information in the cache and no concurrent writes needed, 
+we will use a SQLite DB. We can't put the DB in the DA since it's possible that continuous writes could cause a lot of copies
```
kurtismash: I doubt anyone would understand what "DA" means in this context once we've migrated from DRI.

kurtismash (commented):
Looks good @techncl. A few questions that'll probably be answered during implementation if not now,
- Do we need to consider how to clear this cache?
- Will CC always expect an SQLite DB be present or will this be an optional deployment?
- Would this approach work if we wanted to rebuild the OCFL repo from scratch as we've done previously, or if we wanted to populate a new preservation system with data from CC and "round-trip" it back into CC?

docs/architecture-decision-records/0027-intelligent-caching.md
techncl: It's writing to the database before we send the messages

techncl:
> Looks good @techncl. A few questions that'll probably be answered during implementation if not now,
> 
> * Do we need to consider how to clear this cache? - just by emptying the table
> * Will CC always expect an SQLite DB be present or will this be an optional deployment? - yes, we do expect it to be present.
> * Would this approach work if we wanted to rebuild the OCFL repo from scratch as we've done previously, or if we wanted to populate a new preservation system with data from CC and "round-trip" it back into CC? - yes, we'd need a separate script though.

docs/architecture-decision-records/0027-intelligent-caching.md
kurtismash: Oh, I imagined we'd use S3 events for these messages

kurtismash:
> > Looks good @techncl. A few questions that'll probably be answered during implementation if not now,
> > 
> > * Do we need to consider how to clear this cache? - just by emptying the table
> > * Will CC always expect an SQLite DB be present or will this be an optional deployment? - yes, we do expect it to be present.
> > * Would this approach work if we wanted to rebuild the OCFL repo from scratch as we've done previously, or if we wanted to populate a new preservation system with data from CC and "round-trip" it back into CC? - yes, we'd need a separate script though.

- What about the files within the cache?
- I'd recommend we don't build CC to expect an SQLite off to the side at all times. Make this an optional feature for when it's needed.

One extra, does CC compare the checksum after this fetching from this cache?

MancunianSam (approved):

PR #1090, opened 2026-06-10
DR2-2820 dr2-ingest can StartExecution of Court Doc Preingest Step Fu…

davidkapcsandi:
…nctions


MancunianSam (commented):
Thanks for looking at this. You've passed the missing step function arn into the templatefile but you haven't updated the template file to use this arn so it won't be added to the policy.  If you run `terraform plan` you don't see any changes. 
You need to update the template file `templates/iam_policy/ingest_step_function_policy.json.tpl` to add this variable in with the other preingest variables. 

Could you also add a section to the README in the terraform directory please with a heading like "Adding preingest modules" and put in something about having to update the policy if we ever ad a new module, just so it's documented

davidkapcsandi:
Updated the template file, also added instructions to the README file for future reference.

MancunianSam (approved):

PR #1092, opened 2026-06-12
Minor changes for intelligent caching

MancunianSam:
PurePath().name just returns the file name instead of the full path. We
also don't want the network location in there because this isn't
accounted for in the Preservica database.

I've changed the sys.argv because the first argument is the script name
because Python is fun.

I've added the SET TRANSACTION READ ONLY which will give us some minor
protection against accidentally updating the DRI database.


techncl (approved):
👍

PR #1093, opened 2026-06-17
Changes to state change handler to deal with tape copy confirmation

sparkhi:
Infrastructure:
- New queue 
- update to message states
- configuration 

Lambda:
- tests updated for clarity of existing functionality
- New code to deal with second queue 
- ReadMe updated accordingly


terraform/postingest/postingest.tf
```diff
@@ -1,22 +1,31 @@
...
   six_hours                  = 60 * 60 * 6
   seven_days                 = 60 * 60 * 24 * 7
   messages_visible_threshold = 1000000
   code_deploy_bucket         = var.code_deploy_bucket
+  confirmer_queues = {
+    cc_confirmer = {
+      queue_name = "${var.environment}-dr2-custodial-copy-confirmer"
+    }
+    tape_confirmer = {
+      queue_name = "${var.environment}-dr2-tape-copy-confirmer"
```
kurtismash: Should be `-dr2-postingest-`.

terraform/postingest/postingest.tf
```diff
@@ -1,22 +1,31 @@
...
-  state_change_lambda_name            = "${var.environment}-dr2-${local.state_change_lambda_key}"
-  state_change_lambda_dlq             = "${var.environment}-dr2-postingest-state-change-dlq"
-  resender_lambda_key                 = "postingest-message-resender"
-  resender_lambda_name                = "${var.environment}-dr2-${local.resender_lambda_key}"
-  java_runtime                        = "java21"
-  java_lambda_memory_size             = 512
+  postingest_state_table_name     = "${var.environment}-dr2-postingest-state"
+  postingest_gsi_firstqueued_name = "QueueFirstQueuedIdx"
+  postingest_gsi_lastqueued_name  = "QueueLastQueuedIdx"
+  # custodial_copy_confirmer_queue_name = "${var.environment}-dr2-custodial-copy-confirmer"
```
kurtismash: Why is this commented?

terraform/postingest/postingest.tf
```diff
@@ -1,22 +1,31 @@
...
   postingest_queue_config = [ // Before adding a new queue here, update the state change handler to expect it
-    { "queueAlias" : "CC", "queueOrder" : 1, "queueUrl" : module.dr2_custodial_copy_confirmer_queue.sqs_queue_url }
+    { "queueAlias" : "CC", "queueOrder" : 1, "queueUrl" : module.confirmer_queues["cc_confirmer"].sqs_queue_url },
+    { "queueAlias" : "TC", "queueOrder" : 2, "queueUrl" : module.confirmer_queues["tape_confirmer"].sqs_queue_url }
   ]
   six_hours                  = 60 * 60 * 6
   seven_days                 = 60 * 60 * 24 * 7
   messages_visible_threshold = 1000000
   code_deploy_bucket         = var.code_deploy_bucket
+  confirmer_queues = {
```
kurtismash: This and postingest_queue_config are almost duplicated, is there a better way?

terraform/postingest/postingest.tf
```diff
@@ -67,17 +77,18 @@ module "dr2_custodial_copy_confirmer_queue" {
   delay_seconds                                     = 900
 }
 
-module "cc_confirmer_message_older_than_one_week_alarm" {
+module "confirmer_message_older_than_one_week_alarm" {
   source              = "git::https://github.com/nationalarchives/da-terraform-modules//cloudwatch_alarms"
-  name                = "${local.custodial_copy_confirmer_queue_name}-messages-older-than-one-week-alarm"
+  for_each            = local.confirmer_queues
```
kurtismash: for_each over the queue module instead of the local, reduces the chance of things getting out of sync in future refactors.

terraform/postingest/postingest.tf
```diff
@@ -105,13 +116,13 @@ module "dr2_state_change_lambda" {
 
   policies = {
     "${local.state_change_lambda_name}-policy" = templatefile("${path.module}/templates/policies/state_change_lambda_policy.json.tpl", {
-      custodial_copy_checker_queue_arn = module.dr2_custodial_copy_confirmer_queue.sqs_arn
+      queue_arns                       = jsonencode(flatten([[for _, v in module.confirmer_queues : v.sqs_arn], module.dr2_state_change_lambda_dlq.sqs_arn]))
```
kurtismash: `_,` isn't required here, Terraform defaults to the value when looping over a map.

terraform/postingest/postingest.tf
```diff
@@ -105,13 +116,13 @@ module "dr2_state_change_lambda" {
...
+      custodial_copy_checker_queue_arn = module.confirmer_queues["cc_confirmer"].sqs_arn
```
kurtismash: This is no longer used.

terraform/postingest/postingest.tf
```diff
@@ -67,17 +77,18 @@ module "dr2_custodial_copy_confirmer_queue" {
   delay_seconds                                     = 900
 }
 
-module "cc_confirmer_message_older_than_one_week_alarm" {
+module "confirmer_message_older_than_one_week_alarm" {
   source              = "git::https://github.com/nationalarchives/da-terraform-modules//cloudwatch_alarms"
-  name                = "${local.custodial_copy_confirmer_queue_name}-messages-older-than-one-week-alarm"
+  for_each            = local.confirmer_queues
+  name                = each.value.queue_name
```
kurtismash: What happened to the `-messages-older-than-one-week-alarm` suffix?

terraform/postingest/outputs.tf
```diff
@@ -6,8 +6,10 @@ output "postingest_table_arn" {
   value = module.postingest_state_table.table_arn
 }
 
-output "postingest_confirmer_queue_arn" {
-  value = module.dr2_custodial_copy_confirmer_queue.sqs_arn
+output "postingest_confirmer_queue_arns" {
+  value = {
+    for k, v in module.confirmer_queues : k => v.sqs_arn
```
kurtismash: Why output `k`, it's a meaningless string within this module?

terraform/postingest/outputs.tf
```diff
@@ -18,6 +20,8 @@ output "postingest_resender_lambda_arn" {
   value = module.dr2_message_resender_lambda.lambda_arn
 }
 
-output "cc_confirmer_queue_oldest_message_alarm_arn" {
-  value = module.cc_confirmer_message_older_than_one_week_alarm.cloudwatch_alarm_arn
+output "confirmer_queue_oldest_message_alarm_arns" {
```
kurtismash: "confirmer" is just one of things postingest may do, would advise changing to "postingest_queue_..."

terraform/postingest/outputs.tf
```diff
@@ -18,6 +20,8 @@ output "postingest_resender_lambda_arn" {
...
+  value = {
+    for k, v in module.confirmer_message_older_than_one_week_alarm : k => v.cloudwatch_alarm_arn
```
kurtismash: Why output `k`, it's a meaningless string within this module?

kurtismash (dismissed):
Only reviewed the Terraform. Will this PR include the ADR or will that follow? Do any other docs or the diagram need updated?

terraform/postingest/postingest.tf
sparkhi: custodial-copy-confirmer is an already existing queue without `postingest` in the name, should I change both names?

terraform/postingest/postingest.tf
sparkhi: My bad, I moved the queue name from there to the list and forgot to delete this

terraform/postingest/postingest.tf
kurtismash: No, we'll need to do a migration which is a bigger piece of work - pause ingest, let CC finish, rename, deploy, update the CC worker config. Let's put it in the backlog.

terraform/postingest/postingest.tf
sparkhi: corrected

terraform/postingest/outputs.tf
sparkhi: something like `custodial_copy_user_policy` would need to access it using the key ?

kurtismash: I'd recommend either outputting a list of ARNs or, if we will need to access via a key, using `CC` and `TC` as the keys. `cc_confirmer` is internal to the module and could be changed tomorrow which shouldn't affect things outside of this module.

In the original design of Postingest, the module was to take what is now `local.postingest_queue_config` as a variable and be dynamic. A module should not expect the caller to know things about it's implementation to get values out.

kurtismash (commented):
Worked with @sparkhi to clean up the Terraform implementation. Will leave to another developer to review this and the Scala changes.

scala/lambdas/dynamo-formatters/src/test/scala/uk/gov/nationalarchives/dynamoformatters/DynamoFormattersTest.scala
MancunianSam, line 754: Is it worth adding an assertion here that "result_TC" is equal to `potentialResultTC.get`?

MancunianSam, line 795: Similar thing here. Can you check if "result_TC" is `None`

scala/lambdas/postingest-state-change-handler/src/main/scala/uk/gov/nationalarchives/postingeststatechangehandler/Lambda.scala
```diff
@@ -66,12 +68,13 @@ class Lambda extends LambdaRunner[DynamodbEvent, Unit, Config, Dependencies]:
...
     }
 
-    def updateTableAndSendToSqs(newItem: PostIngestStateTableItem, queue: Queue) =
+    def updateTableAndSendToSqs(newItem: PostIngestStateTableItem, queue: Queue) = {
       updateItem(newItem, queue.queueAlias) >>
         sendMessageToQueue(
           queue.queueUrl,
           OutputQueueMessage(newItem.assetId, newItem.batchId, queue.resultAttrName, newItem.input)
         )
+    }
```
MancunianSam: You don't need these braces here. I guess that was Intellij trying to be helpful.

scala/lambdas/postingest-state-change-handler/src/main/scala/uk/gov/nationalarchives/postingeststatechangehandler/Lambda.scala
```diff
@@ -92,13 +95,14 @@ class Lambda extends LambdaRunner[DynamodbEvent, Unit, Config, Dependencies]:
...
-                val potentialQueue = queues.find(queue => queue.getResult(newItem) != queue.getResult(oldItem))
-
+                val potentialQueue = queues.find(queue => queue.isValid(oldItem, newItem))
                 potentialQueue match {
                   case Some(queue) =>
                     if queue.queueOrder == numOfQueues then deleteItemFromTable(newItem) >> sendOutputMessage(newItem) // new item has met final check; time to delete it from queue
-                    else updateTableAndSendToSqs(newItem, queue) >> sendOutputMessage(newItem, Some(queue.queueAlias))
-                  case _ => IO.unit
+                    else
+                      val nextQueue = queues.find(_.queueOrder == queue.queueOrder + 1).get
```
MancunianSam: Can you put a bit more of a guard against a bad config here. If we have two queues and the order is 1 and 3 then this will fail but the only message we'll get is "Error, None.get"

If you either do something like
```scala
val potentialNextQueue = queues.find(_.queueOrder == queue.queueOrder + 1)
if potentialNextQueue.isDefined then
  val nextQueue = potentialNextQueue.get
   updateTableAndSendToSqs(newItem, nextQueue) >> sendOutputMessage(newItem, Some(nextQueue.queueAlias))
else
  IO.raiseError(new Exception(s"$numOfQueues are in the config but there is no queue with order ${queue.queueOrder + 1}"))
```

Or 
```scala
IO.fromOption(queues.find(_.queueOrder == queue.queueOrder + 1))(new Exception(s"$numOfQueues are in the config but there is no queue with order ${queue.queueOrder + 1}")).flatMap { nextQueue =>
                        updateTableAndSendToSqs(newItem, nextQueue) >> sendOutputMessage(newItem, Some(nextQueue.queueAlias))
                      }
```

MancunianSam (commented):
Looks good thanks. Just a couple of minor bits.

scala/lambdas/postingest-state-change-handler/src/main/scala/uk/gov/nationalarchives/postingeststatechangehandler/Lambda.scala
```diff
@@ -66,20 +70,24 @@ class Lambda extends LambdaRunner[DynamodbEvent, Unit, Config, Dependencies]:
...
 
     def getInsertFibers(queues: List[Queue]) = event.Records
       .filter(_.eventName == EventName.INSERT)
       .parTraverse { record =>
-        val queue1 = queues.find(_.queueOrder == 1).get
-        val newImage = record.dynamodb.newImage.get
-        val processInsertRecord = updateTableAndSendToSqs(newImage, queue1) >> sendOutputMessage(newImage, Some(queue1.queueAlias))
-        processInsertRecord.start
+        val potentialFirstQueue = queues.find(_.queueOrder == 1)
+        if potentialFirstQueue.isDefined then
```
techncl: I don't think that we should do this check here; it's probably better off at the very beginning (the for comprehension at line 129) to do all of the queue checks because here, we are running multiple fibers in parallel.

scala/lambdas/utils/src/main/scala/uk/gov/nationalarchives/utils/PostingestUtils.scala
```diff
@@ -9,18 +9,24 @@ import java.util.UUID
...
 
   sealed trait Queue extends Product {
     def queueAlias: String
     def queueOrder: Int
     def queueUrl: String
     def resultAttrName: String = s"result_$queueAlias"
-    def getResult: PostIngestStateTableItem => Option[String]
+    def getResult(item: PostIngestStateTableItem): Option[String]
+    def isValid(oldItem: PostIngestStateTableItem, newItem: PostIngestStateTableItem): Boolean =
+      oldItem.potentialQueue.contains(queueAlias) && newItem.potentialQueue.contains(queueAlias) && getResult(oldItem) != getResult(newItem)
```
techncl, line 24: I'm confused with the `oldItem.potentialQueue.contains(queueAlias)` part; will the `oldItem` always have a queue alias? In other words, why is it required that the queue exists in the `oldItem?`; could it be blank and the new Item's queue be `CC`?

scala/lambdas/postingest-state-change-handler/src/main/scala/uk/gov/nationalarchives/postingeststatechangehandler/Lambda.scala
```diff
@@ -92,13 +100,22 @@ class Lambda extends LambdaRunner[DynamodbEvent, Unit, Config, Dependencies]:
...
-                val potentialQueue = queues.find(queue => queue.getResult(newItem) != queue.getResult(oldItem))
-
+                val potentialQueue = queues.find(queue => queue.isValid(oldItem, newItem))
                 potentialQueue match {
                   case Some(queue) =>
                     if queue.queueOrder == numOfQueues then deleteItemFromTable(newItem) >> sendOutputMessage(newItem) // new item has met final check; time to delete it from queue
-                    else updateTableAndSendToSqs(newItem, queue) >> sendOutputMessage(newItem, Some(queue.queueAlias))
-                  case _ => IO.unit
+                    else
+                      val potentialNextQueue = queues.find(_.queueOrder == queue.queueOrder + 1)
```
techncl, line 101: how come you're doing this, if a `potentialQueue` exists, that means it is different from the old queue and therefore the "next queue", shouldn't you use `queue` to write to the DB?

scala/lambdas/utils/src/main/scala/uk/gov/nationalarchives/utils/PostingestUtils.scala
```diff
@@ -9,18 +9,24 @@ import java.util.UUID
...
+  case class OutputQueueMessage(assetId: UUID, batchId: String, resultAttrName: String, payload: Json)
 
   sealed trait Queue extends Product {
     def queueAlias: String
     def queueOrder: Int
     def queueUrl: String
     def resultAttrName: String = s"result_$queueAlias"
-    def getResult: PostIngestStateTableItem => Option[String]
+    def getResult(item: PostIngestStateTableItem): Option[String]
+    def isValid(oldItem: PostIngestStateTableItem, newItem: PostIngestStateTableItem): Boolean =
```
techncl: Maybe there is a more descriptive name for this like `resultsDiffer`, `resultsDoNotMatch`?

scala/lambdas/utils/src/main/scala/uk/gov/nationalarchives/utils/PostingestUtils.scala
```diff
@@ -9,18 +9,24 @@ import java.util.UUID
...
+      oldItem.potentialQueue.contains(queueAlias) && newItem.potentialQueue.contains(queueAlias) && getResult(oldItem) != getResult(newItem)
   }
 
-  case class CCQueue(queueAlias: String, queueOrder: Int, queueUrl: String) extends Queue {
-    val getResult: PostIngestStateTableItem => Option[String] = tableItem => tableItem.potentialResultCC
+  case class CCQueue(queueAlias: String, queueOrder: Int = 1, queueUrl: String) extends Queue {
```
techncl: Do we need to add  "= 1" and "=2"? Shouldn't the terraform config be the source of truth?

scala/lambdas/postingest-state-change-handler/src/main/scala/uk/gov/nationalarchives/postingeststatechangehandler/Lambda.scala
```diff
@@ -56,8 +59,9 @@ class Lambda extends LambdaRunner[DynamodbEvent, Unit, Config, Dependencies]:
     def sendOutputMessage(item: PostIngestStateTableItem, newQueueAlias: Option[String] = None): IO[Unit] = {
```
techncl: Now that I think about it, do you think we should change this to be a little more dynamic/error prone? I was thinking that if we added the `messageStatus` to their respective `Queue`s, we don't have to pass in the alias (and would not need to check that it is valid); this would enable us to shorten the first 4 lines into 1

```
    def sendOutputMessage(item: PostIngestStateTableItem, newQueue: Option[Queue] = None): IO[Unit] = {
      val (messageType, messageStatus) = newQueue.map(queue => (IngestUpdate, queue.messageStatus)).getOrElse((IngestComplete, IngestedTape))

      val message = OutputMessage(
        OutputProperties(item.batchId, dependencies.uuidGenerator(), item.potentialCorrelationId, dependencies.instantGenerator(), messageType),
        OutputParameters(item.assetId, messageStatus)
      )
      dependencies.daSnsClient.publish(config.topicArn)(message :: Nil).void
```

terraform/postingest/postingest.tf
```diff
@@ -1,22 +1,30 @@
...
+  state_change_lambda_key         = "postingest-state-change-handler"
+  state_change_lambda_name        = "${var.environment}-dr2-${local.state_change_lambda_key}"
+  state_change_lambda_dlq         = "${var.environment}-dr2-postingest-state-change-dlq"
+  resender_lambda_key             = "postingest-message-resender"
+  resender_lambda_name            = "${var.environment}-dr2-${local.resender_lambda_key}"
+  java_runtime                    = "java21"
+  java_lambda_memory_size         = 512
   postingest_queue_config = [ // Before adding a new queue here, update the state change handler to expect it
-    { "queueAlias" : "CC", "queueOrder" : 1, "queueUrl" : module.dr2_custodial_copy_confirmer_queue.sqs_queue_url }
+    { "queueAlias" : "CC", "queueOrder" : 1, "queue_name" : "${var.environment}-dr2-postingest-custodial-copy-confirmer" },
```
techncl, line 13: Are we accounting for the `queueUrl`/ `queue_name` change in the handler lambda?

techncl (commented):
Looking good! Just a few comments. I've forgotten some of the details of this lambda, so I might be mistaken.

scala/lambdas/utils/src/main/scala/uk/gov/nationalarchives/utils/PostingestUtils.scala
sparkhi: This validation check is only in ModifyFibers. I don't think Old item can be blank. It is basically checking that the modified row has modifications ONLY for the result of that particular queue rather than anything else.

scala/lambdas/postingest-state-change-handler/src/main/scala/uk/gov/nationalarchives/postingeststatechangehandler/Lambda.scala
sparkhi: for a case where, e.g. potentialQueue = CC. We are making sure that this modify event is based on the fact that the change happened in the CC queue, if so, get the next queue (TC) and send the message.

terraform/postingest/postingest.tf
sparkhi: Hmmmm, this one needs a rethink

sparkhi: This is fine, in the environment variables we populate "queueUrl"

scala/lambdas/utils/src/main/scala/uk/gov/nationalarchives/utils/PostingestUtils.scala
techncl: Ah, ok, I confused myself. This `oldItem.potentialQueue.contains(queueAlias) && newItem.potentialQueue.contains(queueAlias)` should always be `true` and `&& getResult(oldItem) != getResult(newItem)` and this is the deciding factor? If it's the case that `oldItem.potentialQueue.contains(queueAlias) && newItem.potentialQueue.contains(queueAlias)`  is `false`, the something seriously wrong has happened, right?

techncl:
I forgot to ask, does the end to end diagram need updating?

terraform/postingest/postingest.tf
```diff
@@ -53,12 +61,13 @@ module "postingest_state_table" {
   point_in_time_recovery_enabled = true
 }
 
-module "dr2_custodial_copy_confirmer_queue" {
+module "confirmer_queues" {
```
techncl: Maybe `dr2_confirmer_queues` for consistency?

terraform/dns_firewall.tf
```diff
@@ -3,16 +3,16 @@ resource "aws_route53_resolver_firewall_domain_list" "allow_domains" {
...
-    "s3.${data.aws_region.current.id}.amazonaws.com.",
-    "s3-r-w.${data.aws_region.current.id}.amazonaws.com.",
-    "dynamodb.${data.aws_region.current.id}.amazonaws.com.",
-    "secretsmanager.${data.aws_region.current.id}.amazonaws.com.",
-    "sts.${data.aws_region.current.id}.amazonaws.com.",
-    "states.${data.aws_region.current.id}.amazonaws.com.",
-    "ssm.${data.aws_region.current.id}.amazonaws.com.",
-    "sqs.${data.aws_region.current.id}.amazonaws.com.",
-    "sns.${data.aws_region.current.id}.amazonaws.com.",
+    "*.s3.${data.aws_region.current.region}.amazonaws.com.",
```
techncl, line 6: Could `data.aws_region.current.region` be saved to a local variable?

techncl (commented):
Just two minor comments

terraform/dns_firewall.tf
sparkhi: I don't think we want to do that. The whole point was to avoid using region from locals. Consolidation in one place should happen as part of this ticket -> https://national-archives.atlassian.net/browse/DR2-2522

techncl (approved):
👍

PR #1126, opened 2026-07-03
Dr2 2856 check terraform drift

davidkapcsandi:


.github/workflows/check-terraform-drift.yml
```diff
@@ -28,7 +28,7 @@ jobs:
         run: |
           terraform init
           terraform workspace select prod
-          export TF_VAR_lambda_code_version=$(gh api /repos/nationalarchives/dr2-ingest/tags | jq -r '.[0].name')
+          export TF_VAR_lambda_code_version=$(git tag --points-at HEAD | grep -E '^v[0-9]+\.[0-9]+\.[0-9]+' | head -1)
```
MancunianSam, line 32: I tried this on another workflow and it's not showing the version. The reason is that the default checkout action doesn't check out the tags.  So in this workflow file, you need to fetch the tags.
```yml
      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
        with:
          ref: 'prod'
          submodules: recursive
          fetch-tags: true
          token: ${{ secrets.WORKFLOW_TOKEN }}
          persist-credentials: false
```

MancunianSam (approved):
👍

PR #1127, opened 2026-07-03
Change flow control config.

MancunianSam:
I put through a single DRI migration ingest and it got stuck on flow control.
It wasn't a TDR or a COURTDOC ingest so it didn't proceed because the default
system had no reserved channels.

This isn't a problem when running the e2e tests, I suspect because there are
other ingests which retrigger the flow control lambda and start it working.

I changed the config in prod to this and it worked.


techncl (approved):
👍

PR #1130, opened 2026-07-07
DR2-2897 Add additional metadata to json file.

MancunianSam:
* Rename `metadata` to `preservicaMetadata` in the DRI migration script. Calling it preservica is fine because it references old Preservica. We're going to delete the existing RG40 and 140 ingests and re-import so there's no need to update existing metadata
* Change the importer code to load the dri metadata and add it to the json file.
* Add an environment variable with the bucket name only for the DRI importer.
* Add extra permissions to allow the lambda to do this.
* Update the tests.


techncl (commented):
Looks good! Would you mind updating the README?

techncl (approved):
👍

PR #1131, opened 2026-07-07
Check latest version in the adhoc script

MancunianSam:
This checks if the script it at the latest version. It will:

* Find the adhoc_ingest.py script and get the parent directory.
* List all files in that directory, filtering out any with _test in the name
* Call the GitHub api to get a list of files in the python/ad-hoc-ingest directory
* Filter out anything with _test in the name
* Iterate through the files from GitHub
* If the file on GitHub is not in the local files, return false
* If it is, download from GitHub and get the md5
* Find the corresponding local file and get the md5
* If they don't match, return false.
* If they do, return true.

This is then checked when the script starts. The user has the option to continue or not.

### DIFF
--- a/python/ad-hoc-ingest/ad_hoc_ingest.py
+++ b/python/ad-hoc-ingest/ad_hoc_ingest.py
@@ -13,8 +13,9 @@ import argument_parser_builder
 import aws_interactions
 import dataset_validator
 import discovery_client
-import metadata_creator
 import message_printer as mp
+import metadata_creator
+import version_check
 
 
 def validate_arguments(args):
@@ -143,6 +144,10 @@ def is_folder_writable(output_folder):
 
 
 def main():
+    if not version_check.is_latest_version():
+        version_check_response = input("Adhoc ingest is not at the latest version. Do you want to continue? y/n")
+        if version_check_response.lower() not in ["y", "yes"]:
+            sys.exit(0)
     args = argument_parser_builder.build().parse_args()
     validate_arguments(args)
 
@@ -171,3 +176,4 @@ def main():
 
 if __name__ == "__main__":
     main()
+
--- a/python/ad-hoc-ingest/ad_hoc_ingest.py
+++ b/python/ad-hoc-ingest/ad_hoc_ingest.py
@@ -146,7 +146,8 @@ def is_folder_writable(output_folder):
 def main():
     if not version_check.is_latest_version():
         version_check_response = input("Adhoc ingest is not at the latest version. Do you want to continue? y/n")
-        if version_check_response.lower() not in ["y", "yes"]:
+        if version_check_response.lower().strip() not in ["y", "yes"]:
+            mp.print_message("Exiting")
             sys.exit(0)
     args = argument_parser_builder.build().parse_args()
     validate_arguments(args)
--- a/python/ad-hoc-ingest/ad_hoc_ingest_test.py
+++ b/python/ad-hoc-ingest/ad_hoc_ingest_test.py
@@ -13,7 +13,7 @@ import ad_hoc_ingest
 
 class Test(TestCase):
     def setUp(self):
-        self.test_dir = tempfile.mkdtemp()
+        self.test_dir = os.path.realpath(tempfile.mkdtemp())
 
     def tearDown(self):
         shutil.rmtree(self.test_dir)
@@ -214,6 +214,49 @@ JS 8,someRecordId,someFileId,"Description of Kew, Richmond, London",JS-8-3.pdf,3
         account_number_after_refresh = ad_hoc_ingest.get_account_number()
         mock_refresh_session.assert_called_once()
         self.assertEqual("987654321", account_number_after_refresh)
+
+    @patch("version_check.is_latest_version", return_value=False)
+    @patch("builtins.input", return_value="n")
+    def test_should_exit_when_not_at_latest_version_and_user_declines_to_continue(self, mock_input, _):
+        with self.assertRaises(SystemExit) as exc:
+            ad_hoc_ingest.main()
+        self.assertEqual(0, exc.exception.code)
+        mock_input.assert_called_once_with("Adhoc ingest is not at the latest version. Do you want to continue? y/n")
+
+    @patch("version_check.is_latest_version", return_value=False)
+    @patch("builtins.input", return_value="no")
+    def test_should_exit_when_not_at_latest_version_and_user_types_no(self, _, __):
+        with self.assertRaises(SystemExit) as exc:
+            ad_hoc_ingest.main()
+        self.assertEqual(0, exc.exception.code)
+
+    @patch("argument_parser_builder.build")
+    @patch("version_check.is_latest_version", return_value=False)
+    @patch("builtins.input", return_value="y")
+    def test_should_continue_when_not_at_latest_version_and_user_confirms_with_y(self, _, __, mock_build):
+        mock_build.side_effect = Exception("Version check passed")
+        with self.assertRaises(Exception) as e:
+            ad_hoc_ingest.main()
+        self.assertEqual("Version check passed", str(e.exception))
+
+    @patch("argument_parser_builder.build")
+    @patch("version_check.is_latest_version", return_value=False)
+    @patch("builtins.input", return_value="yes")
+    def test_should_continue_when_not_at_latest_version_and_user_confirms_with_yes(self, _, __, mock_build):
+        mock_build.side_effect = Exception("Version check passed")
+        with self.assertRaises(Exception) as e:
+            ad_hoc_ingest.main()
+        self.assertEqual("Version check passed", str(e.exception))
+
+    @patch("argument_parser_builder.build")
+    @patch("version_check.is_latest_version", return_value=True)
+    def test_should_skip_version_prompt_when_already_at_latest_version(self, _, mock_build):
+        mock_build.side_effect = Exception("Version check passed")
+        with patch("builtins.input") as mock_input:
+            with self.assertRaises(Exception) as e:
+                ad_hoc_ingest.main()
+        self.assertEqual("Version check passed", str(e.exception))
+        mock_input.assert_not_called()
 
     @patch("aws_interactions.get_account_number")
     @patch("aws_interactions.refresh_session")
--- a/python/ad-hoc-ingest/ad_hoc_ingest_test.py
+++ b/python/ad-hoc-ingest/ad_hoc_ingest_test.py
@@ -241,7 +241,25 @@ JS 8,someRecordId,someFileId,"Description of Kew, Richmond, London",JS-8-3.pdf,3
 
     @patch("argument_parser_builder.build")
     @patch("version_check.is_latest_version", return_value=False)
+    @patch("builtins.input", return_value="Y")
+    def test_should_continue_when_not_at_latest_version_and_user_confirms_with_y_uppercase(self, _, __, mock_build):
+        mock_build.side_effect = Exception("Version check passed")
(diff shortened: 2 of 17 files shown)


python/ad-hoc-ingest/ad_hoc_ingest.py
```diff
@@ -142,7 +145,31 @@ def is_folder_writable(output_folder):
...
+            version = v.read()
+    else:
+        raise Exception("Cannot find version file")
+
+    mp.print_message(f"Ad hoc ingest version {version}")
+    github_url = "https://raw.githubusercontent.com/nationalarchives/dr2-ingest/refs/heads/main/python/ad-hoc-ingest/version"
+    try:
+        with urlopen(github_url) as gh:
+            latest_version = gh.read().decode("utf-8").strip()
+            if latest_version.split(".")[2] > version.split(".")[2]:
```
techncl: 1. Should we convert these values to Ints as `"1000" > "999" == false`?
2. Are we assuming that the version is always going to start with "0.0"?

python/ad-hoc-ingest/ad_hoc_ingest.py
```diff
@@ -142,7 +145,31 @@ def is_folder_writable(output_folder):
...
+    if path.exists("version"):
+        with open("version") as v:
+            version = v.read().strip()
+    elif path.exists("ad-hoc-ingest/version"):
+        with open("ad-hoc-ingest/version") as v:
+            version = v.read()
+    else:
+        raise Exception("Cannot find version file")
+
+    mp.print_message(f"Ad hoc ingest version {version}")
```
techncl: Maybe f"Current Ad hoc ingest version {version}"?

python/ad-hoc-ingest/ad_hoc_ingest.py
```diff
@@ -143,6 +144,10 @@ def is_folder_writable(output_folder):
 
 
 def main():
+    if not version_check.is_latest_version():
+        version_check_response = input("Adhoc ingest is not at the latest version. Do you want to continue? y/n")
```
techncl, line 148: 1. "on" instead of "at"
2. Should we link them to the latest version?

python/ad-hoc-ingest/ad_hoc_ingest.py
```diff
@@ -143,6 +144,10 @@ def is_folder_writable(output_folder):
...
+        if version_check_response.lower() not in ["y", "yes"]:
```
techncl: Should we add ".strip()" to this?

python/ad-hoc-ingest/ad_hoc_ingest.py
```diff
@@ -143,6 +144,10 @@ def is_folder_writable(output_folder):
...
+            sys.exit(0)
```
techncl, line 151: Should we add a message telling them that it's quit because I'm thinking, if someone types a "u" instead of a "y" or something and clicks enter, the script will just exit

python/ad-hoc-ingest/ad_hoc_ingest_test.py
```diff
@@ -215,6 +215,49 @@ def test_should_call_refresh_session_when_a_client_error_is_thrown_when_getting_
...
+    @patch("version_check.is_latest_version", return_value=False)
+    @patch("builtins.input", return_value="no")
+    def test_should_exit_when_not_at_latest_version_and_user_types_no(self, _, __):
+        with self.assertRaises(SystemExit) as exc:
+            ad_hoc_ingest.main()
+        self.assertEqual(0, exc.exception.code)
+
+    @patch("argument_parser_builder.build")
+    @patch("version_check.is_latest_version", return_value=False)
+    @patch("builtins.input", return_value="y")
```
techncl, line 235: Is it worth capitalising this "y" as it will also test the ".lower()" that you added?

python/ad-hoc-ingest/ad_hoc_ingest_test.py
```diff
@@ -215,6 +215,49 @@ def test_should_call_refresh_session_when_a_client_error_is_thrown_when_getting_
...
+    def test_should_continue_when_not_at_latest_version_and_user_confirms_with_y(self, _, __, mock_build):
+        mock_build.side_effect = Exception("Version check passed")
+        with self.assertRaises(Exception) as e:
+            ad_hoc_ingest.main()
+        self.assertEqual("Version check passed", str(e.exception))
+
+    @patch("argument_parser_builder.build")
+    @patch("version_check.is_latest_version", return_value=False)
+    @patch("builtins.input", return_value="yes")
```
techncl, line 253: Is it worth capitalising this "yes" as it will also test the ".lower()" that you added?

python/ad-hoc-ingest/version_check.py
```diff
@@ -0,0 +1,49 @@
...
+        return response.read()
+
+
+def md5_of_bytes(data: bytes) -> str:
+    return hashlib.md5(data).hexdigest()
+
+
+def fetch_remote_files() -> list[dict]:
+    entries = fetch_json(REPO_API_URL)
+    return [e for e in entries if e["type"] == "file" and "_test" not in e["name"]]
```
techncl, line 29: Should we exclude the `README`s and `requirements.txt` files

python/ad-hoc-ingest/version_check_test.py
```diff
@@ -0,0 +1,158 @@
...
+
+    # --- fetch_json ---
+
+    @patch("urllib.request.urlopen")
+    def test_fetch_json_should_return_parsed_json(self, mock_urlopen):
+        payload = [{"name": "file.py", "type": "file"}]
+        mock_urlopen.return_value = self.mock_urlopen(json.dumps(payload))
+        result = version_check.fetch_json("https://example.com/api")
+        self.assertEqual(payload, result)
+        mock_urlopen.assert_called_once()
```
techncl, line 42: Should we check that `urllib.request.urlopen` was passed the correct args?

python/ad-hoc-ingest/version_check_test.py
```diff
@@ -0,0 +1,158 @@
...
+
+    @patch("urllib.request.urlopen")
+    def test_fetch_bytes_should_return_raw_bytes(self, mock_urlopen):
+        expected = b"raw file content"
+        mock_urlopen.return_value = self.mock_urlopen(expected)
+        result = version_check.fetch_bytes("https://example.com/file.py")
+        self.assertEqual(expected, result)
```
techncl, line 51: Should we check that urllib.request.urlopen was passed the correct args?

python/ad-hoc-ingest/version_check_test.py
```diff
@@ -0,0 +1,158 @@
...
+        self.assertNotIn("some_dir", names)
+
+    @patch("version_check.fetch_json")
+    def test_fetch_remote_files_should_return_empty_list_when_no_matching_entries(self, mock_fetch_json):
+        mock_fetch_json.return_value = [
+            {"name": "dataset_validator_test.py", "type": "file"},
+            {"name": "subdir", "type": "dir"},
+        ]
+        result = version_check.fetch_remote_files()
+        self.assertEqual([], result)
```
techncl, line 90: 1. Should we test that the "fetch_json" is passed the correct url?
2. Do we need to pass in the url, would it not just be in the function or it refers to it directly, since it is global?

python/ad-hoc-ingest/version_check.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+def is_latest_version() -> bool:
+    remote_files = fetch_remote_files()
```
techncl, line 33: Is it worth asserting that `remote_files` is non-empty, just in case something weird happens? I saw that we have a test for empty remote files but I couldn't think of a positive/expected reason why there would not be any remote files

python/ad-hoc-ingest/version_check_test.py
```diff
@@ -0,0 +1,158 @@
...
+            {"name": "version_check.py", "download_url": "https://example.com/version_check.py"},
+        ]
+        mock_fetch_bytes.return_value = content
+
+        with patch("os.walk") as mock_walk, patch("os.listdir") as mock_listdir:
+            mock_walk.return_value = [(self.test_dir, [], ["ad_hoc_ingest.py", "version_check.py"])]
+            mock_listdir.return_value = ["ad_hoc_ingest.py", "version_check.py"]
+            result = version_check.is_latest_version()
+
+        self.assertTrue(result)
```
techncl, line 111: 1. Should we confirm that `fetch_remote_files` was called?
2. Same question for all of the other "is_latest_version" tests

python/ad-hoc-ingest/version_check.py
techncl, line 33: Is it worth asserting that remote_files is non-empty, just in case something weird happens? I saw that we have a test for empty remote files but I couldn't think of a positive/expected reason why there would not be any remote files.

techncl (commented):
Looks good! Just some minor comments. Also, should we update the README with this?

python/ad-hoc-ingest/ad_hoc_ingest.py
MancunianSam: This is not there any more. We're doing it a different way.

python/ad-hoc-ingest/ad_hoc_ingest.py
MancunianSam: Same here, this has gone now.

python/ad-hoc-ingest/ad_hoc_ingest.py
techncl: That's weird; it must have cached the old comments on my old review even though I deleted them.

python/ad-hoc-ingest/ad_hoc_ingest.py
MancunianSam: 1. Internet consensus seems to be that "at" is more commonly used for a software version.
2. Not now, I don't know how people are getting the code, I don't want to make things confusing.

python/ad-hoc-ingest/ad_hoc_ingest.py
MancunianSam: Yeah ok.

python/ad-hoc-ingest/ad_hoc_ingest_test.py
MancunianSam: I've added another test that uses `Y` as well as this one.

python/ad-hoc-ingest/ad_hoc_ingest_test.py
MancunianSam: I've added another test for `YES` as well.

python/ad-hoc-ingest/version_check.py
MancunianSam: No. `README`s are rarely updated on their own and the `requirements.txt` changes might be version bumps for security fixes.

python/ad-hoc-ingest/version_check_test.py
MancunianSam: I've added a test that checks it calls the right url. I don't think we need to pass it in, it's fine as a global variable.

python/ad-hoc-ingest/version_check.py
MancunianSam: Yeah why not. If there's no files then something has gone very wrong. I've added an assertion and changed the test to look for the error.

MancunianSam: This is a duplicate comment. I've added the check.

techncl (approved):
👍

PR #1133, opened 2026-07-08
Implement lambda + SQS between postingest-state DDB and Change Handler

techncl:


terraform/postingest/postingest.tf
```diff
@@ -1,15 +1,20 @@
...
+  state_change_lambda_name                   = "${var.environment}-dr2-${local.state_change_lambda_key}"
+  state_change_lambda_dlq                    = "${var.environment}-dr2-postingest-state-change-dlq"
+  resender_lambda_key                        = "postingest-message-resender"
+  resender_lambda_name                       = "${var.environment}-dr2-${local.resender_lambda_key}"
+  java_runtime                               = "java21"
+  java_lambda_memory_size                    = 512
+  python_timeout_seconds                     = 30
+  python_runtime                             = "python3.14"
+  python_lambda_memory_size                  = 128
+  send_to_state_change_ddb_queue_lambda_name = "${var.environment}-dr2-send_to_state_change_ddb_queue"
```
kurtismash: Our naming convention should be `env-dr2-postingest-` for this module.

Why a mix of hyphens and underscores?

terraform/postingest/postingest.tf
```diff
@@ -1,15 +1,20 @@
...
-  state_change_lambda_dlq             = "${var.environment}-dr2-postingest-state-change-dlq"
-  resender_lambda_key                 = "postingest-message-resender"
-  resender_lambda_name                = "${var.environment}-dr2-${local.resender_lambda_key}"
-  java_runtime                        = "java21"
-  java_lambda_memory_size             = 512
+  postingest_state_table_name                = "${var.environment}-dr2-postingest-state"
+  postingest_gsi_firstqueued_name            = "QueueFirstQueuedIdx"
+  postingest_gsi_lastqueued_name             = "QueueLastQueuedIdx"
+  custodial_copy_confirmer_queue_name        = "${var.environment}-dr2-custodial-copy-confirmer"
+  state_change_ddb_queue_name                = "${var.environment}-dr2-state-change-handler"
```
kurtismash: Our naming convention should be `env-dr2-postingest-` for this module.

terraform/postingest/postingest.tf
```diff
@@ -111,18 +136,22 @@ module "dr2_state_change_lambda" {
       account_id                       = data.aws_caller_identity.current.account_id
       lambda_name                      = local.state_change_lambda_name
       dynamo_db_postingest_stream_arn  = module.postingest_state_table.stream_arn
-      state_change_dlq_arn             = module.dr2_state_change_lambda_dlq.sqs_arn
+      state_change_dlq_arn             = module.dr2_state_change_ddb_queue.dlq_sqs_arn
```
kurtismash: We shouldn't need this anymore, as the Lambda is now invoked by SQS we can let the SQS Event Source Mapping handle this.

terraform/postingest/postingest.tf
```diff
@@ -83,19 +88,39 @@ module "cc_confirmer_message_older_than_one_week_alarm" {
   threshold = local.seven_days
 }
 
-module "dr2_state_change_lambda_dlq" {
+module "dr2_state_change_ddb_queue" {
   source     = "git::https://github.com/nationalarchives/da-terraform-modules//sqs"
-  queue_name = local.state_change_lambda_dlq
-  sqs_policy = templatefile("./templates/sqs/lambda_sqs_dlq_policy.json.tpl", {
+  queue_name = local.state_change_ddb_queue_name
+  sqs_policy = templatefile("./templates/sqs/sqs_access_policy.json.tpl", {
```
kurtismash: Does this file exist?

kurtismash (dismissed):
Only reviewed the Python and Terraform.

terraform/postingest/postingest.tf
techncl: Ah, yes. I just copied and pasted from the lambda and forgot to check

terraform/postingest/postingest.tf
techncl: Yes, [here](https://github.com/nationalarchives/dr2-ingest/blob/main/terraform/templates/sqs/sqs_access_policy.json.tpl)

kurtismash: Oh, it's outside the module. Let's not do that, modules shouldn't be dependent on information that isn't inside of them or directly passed into them.

We should also consider if this policy is enforcing least privilege.

scala/lambdas/postingest-state-change-handler/src/main/scala/uk/gov/nationalarchives/postingeststatechangehandler/Lambda.scala
```diff
@@ -121,11 +130,17 @@ class Lambda extends LambdaRunner[DynamodbEvent, Unit, Config, Dependencies]:
...
       allResults <- (insert ++ modify).parTraverse(_.join)
-      _ <- allResults.traverse {
-        case Outcome.Errored(e) => IO.raiseError(e)
-        case _                  => IO.unit
+      batchItemFailures <- allResults.traverseFilter {
+        case Outcome.Errored(e) =>
+          IO.pure {
+            e match {
+              case e: StateChangeException => Some(BatchItemFailure(e.sequenceNumber))
+              case _                       => None
```
MancunianSam: Doesn't this mean that if we get a failed outcome on this thread, we don't send the failure in the response which means that this message will be marked as successful?

This is unlikely because you're catching the exceptions and converting them further up but is it worth either re-raising it here or just removing this line and letting us get a `MatchError` if this ever does happen.

python/lambdas/send-to-state-change-ddb-queue/README.md
```diff
@@ -0,0 +1,149 @@
+# DR2 Send to Post-Ingest State Change DDB Queue
+
+## Input
+
```
sparkhi, line 25: `item` instead of `entry` in DynamoDB may appear more correct

python/lambdas/send-to-state-change-ddb-queue/README.md
```diff
@@ -0,0 +1,149 @@
...
+The lambda is triggered via a DynamoDB stream whenever an entry in DynamoDB is changed ("MODIFY"),
+removed or added ("INSERT").
+
```
sparkhi, line 28: You could add one line stating that the remove event is ignored

python/lambdas/send-to-state-change-ddb-queue/README.md
```diff
@@ -0,0 +1,149 @@
...
+## Output
+
+The lambda doesn't return anything, but it passes the event on to an SQS `QUEUE_URL`
+
+
+## Steps
+
+1. Send the `event` recevied to the `QUEUE_URL`
+
+[Link to the infrastructure code](https://github.com/nationalarchives/dp-terraform-environments)
```
sparkhi: this is probably a result of copy+paste. Most of our ReadMe files have this link to the archived repository (pre mono repo?) I think it should be (https://github.com/nationalarchives/dr2-ingest/tree/main/terraform)

scala/lambdas/postingest-state-change-handler/src/main/scala/uk/gov/nationalarchives/postingeststatechangehandler/Lambda.scala
```diff
@@ -216,4 +231,6 @@ object Lambda:
 
   case class DynamodbStreamRecord(eventName: EventName, dynamodb: StreamRecord)
 
-  case class StreamRecord(keys: Option[PostIngestStatePrimaryKey], oldImage: Option[PostIngestStateTableItem], newImage: Option[PostIngestStateTableItem])
+  case class StreamRecord(keys: Option[PostIngestStatePrimaryKey], oldImage: Option[PostIngestStateTableItem], newImage: Option[PostIngestStateTableItem], sequenceNumber: String)
+
+  case class StateChangeException(message: String, sequenceNumber: String) extends Exception(message)
```
sparkhi, line 250: would it be better to extend `RuntimeException` ?

sparkhi (commented):
A few minor comments

scala/lambdas/postingest-state-change-handler/src/main/scala/uk/gov/nationalarchives/postingeststatechangehandler/Lambda.scala
techncl: I don't think so as we want to capture all Exceptions and send them back onto the queue

MancunianSam (dismissed):

python/lambdas/send-to-state-change-ddb-queue/send_to_state_change_ddb_queue.py
```diff
@@ -0,0 +1,13 @@
...
+import boto3
+import os
+
+sqs_client = boto3.client("sqs")
+
+
+def lambda_handler(event, context):
+    sqs_client.send_message(
+        QueueUrl=os.environ["QUEUE_URL"],
+        MessageBody=json.dumps(event),
```
kurtismash: This will send a single SQS message that contains many DynamoDB updates, recreating the issue we've had where a rogue update to a single asset causes many assets to get stuck in Postingest. 

In Terraform, you've not set a `batch_size` in the `dynamo_stream_config` of `modue.dr2_state_change_ddb_queue_lambda`, therefore the event passed into the Lambda [will contain up to 100 item updates](https://github.com/nationalarchives/da-terraform-modules/blob/a5ae72cecded5db666301dbc6d6a16bf9019be42/lambda/variables.tf#L156) and become the message.

Your setting of `sqs_queue_mapping_batch_size = 10` and `sqs_report_batch_item_failures = true` partially mitigates this, but as each message could contain 100 items we'd still see 100 items fail to process due to 1 item failure.

terraform/postingest/postingest.tf
kurtismash: Should this now be the new SQS Queue ARN?

terraform/postingest/postingest.tf
```diff
@@ -133,10 +149,14 @@ module "dr2_state_change_lambda" {
...
   runtime     = local.java_runtime
-  dynamo_stream_config = {
-    stream_arn             = module.postingest_state_table.stream_arn
-    dead_letter_target_arn = module.dr2_state_change_lambda_dlq.sqs_arn
-  }
+
+  sqs_queue_mapping_batch_size   = 10
+  sqs_report_batch_item_failures = true
+  lambda_sqs_queue_mappings = [{
+    sqs_queue_arn         = "arn:aws:sqs:eu-west-2:${data.aws_caller_identity.current.account_id}:${local.state_change_ddb_queue_name}"
```
kurtismash, line 156: Why can't this be the SQS ARN from `module.dr2_state_change_ddb_queue`?

kurtismash (changes requested):

scala/lambdas/postingest-state-change-handler/src/main/scala/uk/gov/nationalarchives/postingeststatechangehandler/Lambda.scala
sparkhi: You are catching all exceptions and wrapping them in the StateChangeException (line 88) which is fine, however the Exception that is defined here does not have to be checked exception hence the suggestion of extending from RuntimeException

python/lambdas/send-to-state-change-ddb-queue/send_to_state_change_ddb_queue.py
techncl: So should the `batch_size` be set to 1?

kurtismash: I'd recommend wrapping the `send_message` call in a for loop over the `event["Records"]` array, that will send a message for each item update regardless of `batch_size`.

terraform/postingest/postingest.tf
techncl: Because I was getting [an error](https://github.com/nationalarchives/dr2-ingest/actions/runs/29758932898/job/88408372135) when `terraform plan` was run as terraform doesn't know what this value is until after the `apply` so adding this explicitly fixed it.

kurtismash (approved):

PR #1137, opened 2026-07-13
DR2-2903 Update Ids.

MancunianSam:
TDR have added an AssetId field although not all transfers will have it just yet.
At some point in the future they will stop using the UUID field.
So both of these need to be optional so TDR can change things without breaking our package builder.


techncl (commented):
Looks good!  Would you mind updating the README also?

techncl (approved):
👍

PR #1138, opened 2026-07-16
DR2-2898 Log ids before upload

MancunianSam:
We are geting files left behind when the judgments are too large and the lambda runs out of memory.

I think this is because some are being copied over before they time out. We can't tell what they are at the moment
because the key we upload them to the raw cache bucket with is a random uuid.

This will log it as they're uploaded so if we get another OOM error in the courtdoc package builder, we can see which files it has copied over and
delete them manually.


techncl (approved):
👍

PR #1139, opened 2026-07-16
DR2-2896 E2E tests improvements

MancunianSam:
There is now an extra step for the happy path which polls the lock table for the ids and adds the group ids to a list
Then after we have got the messages, we check that the step function has completed and that the expected files have been deleted.

I've also added a single ~20Mb file to the ingests to test multipart copy


kotlin/e2e-tests/lib/src/main/kotlin/uk/gov/nationalarchives/lib/IngestUtils.kt
```diff
@@ -50,11 +61,37 @@ class IngestUtils(
...
     private val sfnClient: SfnClient,
     private val config: Config,
-    private val assetIds: MutableList<UUID>
+    private val assetIds: MutableList<UUID>,
+    private val groupIds: MutableSet<String>
 ) {
     private val completeStatus = "Asset has been written to custodial copy disk."
     private val failedStatus = "There has been an error ingesting the asset."
 
+    suspend fun waitForEntriesInLockTable(timeout: Long = 15 * 60 * 1000): MutableSet<String> {
```
techncl: Should we use `15.minutes`?

kotlin/e2e-tests/lib/src/main/kotlin/uk/gov/nationalarchives/lib/IngestUtils.kt
```diff
@@ -50,11 +61,37 @@ class IngestUtils(
...
+        val pollInterval = 30 * 1000
```
techncl, line 73: 1. should this be 30.minutes?
2. could this be global and used on line 149 and 183?

kotlin/e2e-tests/lib/src/main/kotlin/uk/gov/nationalarchives/lib/IngestUtils.kt
```diff
@@ -50,11 +61,37 @@ class IngestUtils(
...
+        val pendingAssetIds = assetIds.map { it.toString() }.toMutableSet()
+        val tableName = config.getString("lockTable")
+        try {
+            withTimeout(timeout.milliseconds) {
+                while (pendingAssetIds.isNotEmpty()) {
+                    val rows = pendingAssetIds
+                        .chunked(100)
+                        .flatMap { assetIdsChunk -> getBatchRows(tableName, assetIdsChunk) }
+                    rows.forEach { row ->
+                        row["assetId"]?.asS()?.let { pendingAssetIds.remove(it) }
```
techncl, line 83: So from what I understand, `?` are like Option is Scala, so if `row["assetId"]?.asS()?` yields `null`, shouldn't we throw an error otherwise it will continue to loop?

kotlin/e2e-tests/lib/src/main/kotlin/uk/gov/nationalarchives/lib/IngestUtils.kt
```diff
@@ -50,11 +61,37 @@ class IngestUtils(
...
+                        row["groupId"]?.asS()?.takeIf { it.isNotBlank() }?.let { groupIds.add(it) }
```
techncl, line 84: Should we only do this if an `assetId` exists?

kotlin/e2e-tests/lib/src/main/kotlin/uk/gov/nationalarchives/lib/IngestUtils.kt
```diff
@@ -92,6 +129,70 @@ class IngestUtils(
...
 
+    private suspend fun describeStepFunction(groupId: String): DescribeExecutionResponse {
+        val stepFunctionName = "${groupId}_0"
+        val describeRequest = DescribeExecutionRequest {
+            executionArn = "arn:aws:states:eu-west-2:${config.getString("accountNumber")}:execution:${config.getString("ingestSfnName")}:$stepFunctionName"
+        }
+        return sfnClient.describeExecution(describeRequest)
+    }
+
+    suspend fun checkStepFunctionCompletes(timeout: Long = 15 * 60 * 1000) {
```
techncl: Should we use `15.minutes`?

kotlin/e2e-tests/lib/src/main/kotlin/uk/gov/nationalarchives/lib/IngestUtils.kt
```diff
@@ -92,6 +129,70 @@ class IngestUtils(
...
+                            if (!notDeletedFiles.isNullOrEmpty()) {
+                                delay(30.seconds)
+                            }
+                        } while (!notDeletedFiles.isNullOrEmpty())
+                    }
+                }
+            } catch (_: TimeoutCancellationException) {
+                throw TimeoutException("Timed out waiting for step function completion for group id $groupId")
+            }
+
```
techncl: There are two new lines here

kotlin/e2e-tests/lib/src/main/kotlin/uk/gov/nationalarchives/lib/IngestUtils.kt
```diff
@@ -103,13 +204,16 @@ class IngestUtils(
         val invalidChecksumValue = "e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855"
         assetIds.addAll(List<UUID>(numberOfFiles) { UUID.randomUUID() })
         coroutineScope {
-            assetIds.map {
+            assetIds.mapIndexed { index, assetId ->
                 launch {
+                    val body = if (index == 0) assetId.toString().repeat(600000)
```
techncl: Could we add a comment as to why it's repeating this 600,000 times?

kotlin/e2e-tests/lib/src/test/kotlin/uk/gov/nationalarchives/lib/IngestUtilsTest.kt
```diff
@@ -305,6 +308,108 @@ class IngestUtilsTest {
...
+            listOf("key"),
+            groupIds,
+            ExecutionStatus.Succeeded
+        )
+        utils.checkStepFunctionCompletes()
+    }
+
+    @Test
+    fun checkStepFunctionCompletesReturnsErrorIfStatusNotSucceeded(): Unit = runBlocking {
+        val metadataPackageUri = URI("s3://bucket/metadata")
```
techncl, line 355: Is it worth making these 4 or 5 lines global variables?

techncl (commented):
Looks good! Just some minor comments.

kotlin/e2e-tests/lib/src/main/kotlin/uk/gov/nationalarchives/lib/IngestUtils.kt
MancunianSam: 1. No, 30 seconds, the lock table entries should only take a few minutes to appear.
2. I don't think so. We might want to tweak some of the others at some point and it's clearer having the time directly where it's used.

techncl: Sorry, copy and paste error; I meant "seconds" i.e. should we do `30.seconds` instead?

kotlin/e2e-tests/lib/src/main/kotlin/uk/gov/nationalarchives/lib/IngestUtils.kt
MancunianSam: I think if the `assetId` isn't there then we've got a fairly serious problem elsewhere so it's unlikely. It's not worth making the code more complex here for something unlikely and it will fail eventually after it times out. Also, we're querying by assetId so it can't not exist.

kotlin/e2e-tests/lib/src/main/kotlin/uk/gov/nationalarchives/lib/IngestUtils.kt
MancunianSam: We're querying by the assetId so it will be here. And even if something really unlikely happens, the tests will still fail so I think it's fine as is.

kotlin/e2e-tests/lib/src/test/kotlin/uk/gov/nationalarchives/lib/IngestUtilsTest.kt
MancunianSam: No, I would like to keep all the inputs for a test inside the test, even if there's a bit of repetition. It's difficult to read if there's shared state everywhere.

techncl (approved):
👍

PR #1140, opened 2026-07-16
Dr2 2855 ad hoc ingest should have the ability to supply own descriptions

sparkhi:
Description column is read from the input file
if description column is included in input, then we do not go to Discovery, but use the supplied value

There is ability to supply digitalAssetSource as a CLI param


python/ad-hoc-ingest/ad_hoc_ingest.py
```diff
@@ -165,12 +164,17 @@ def main():
     except Exception as e:
         raise Exception(f"Inputs supplied to the process are invalid, please fix errors before continuing: {e}")
 
-    is_discovery_available = discovery_client.is_discovery_api_reachable()
-    if not is_discovery_available:
-        mp.print_message("Discovery API is not available for getting metadata information, terminating process")
-        sys.exit(1)
+    process_upload(data_set, args, is_valid)
+
+def process_upload(data_set, args, is_dataset_valid):
```
techncl: Because `process_upload` is doing the Discovery thing as well, maybe it might be better/cleaner/clearer/less nested to:
   1. rename this function to something like `check_discovery_availability`
   2. return `description_override`
   3. put lines 177-180 under line 1672.  Maybe this function could be moved to just above `upload_files_to_ingest_bucket` so that main is the first function and so that it flows better when reading it?

You also, wouldn't need to pass in `args` or `is_dataset_valid`

python/ad-hoc-ingest/ad_hoc_ingest.py
```diff
@@ -165,12 +164,17 @@ def main():
...
+    description_override = "description" in data_set.columns
```
techncl, line 170: 1. Maybe "override_description" so that it sounds more like a question?
2. What if the "description" column is uppercased or has a leading/trailing space?

python/ad-hoc-ingest/ad_hoc_ingest.py
```diff
@@ -165,12 +164,17 @@ def main():
...
+    if not description_override:
```
techncl, line 171: Do you think it's worth, if a description is provided, logging something like ""description" column detected in CSV; skipping Discovery lookup"?

python/ad-hoc-ingest/ad_hoc_ingest.py
```diff
@@ -165,12 +164,17 @@ def main():
...
+        is_discovery_available = discovery_client.is_discovery_api_reachable()
```
techncl: Do we need this intermediate variable since the method of the client is quite descriptive and we only use the variable once?

python/ad-hoc-ingest/README.md
techncl, line 25: Could you add the caveat here that this only happens if a description isn't provided?

python/ad-hoc-ingest/metadata_creator.py
```diff
@@ -4,34 +4,40 @@
...
 
-    if not collection_info.title and not collection_info.description:
-        raise Exception(f"Title and Description both are empty for '{catalog_ref}', unable to proceed with this record")
+        if not collection_info.title and not collection_info.description:
+            raise Exception(f"Title and Description both are empty for '{catalog_ref}', unable to proceed with this record")
 
-    description_to_use = collection_info.title if collection_info.title is not None else collection_info.description
+        description_to_use = collection_info.title if collection_info.title is not None else collection_info.description
+    else:
+        former_ref_dept = None
```
techncl: Because of line 33, why not set this to an empty string?

python/ad-hoc-ingest/metadata_creator.py
```diff
@@ -4,34 +4,40 @@
...
+        former_ref_tna = None
```
techncl: Because of line 33, why not set this to an empty string?

python/ad-hoc-ingest/metadata_creator.py
```diff
@@ -4,34 +4,40 @@
...
-        "Filename": get_filename_from_cross_platform_path(file_path),
-        "FileReference": catalog_ref.removeprefix(f"{series}").strip().removeprefix("/").strip(),
-        "ClientSideOriginalFilepath": file_path,
-        "IAID": discovery_iaid,
-    }
-    former_ref_dept = former_references.formerRefDept
-    metadata["formerRefDept"] = "" if former_ref_dept is None else former_ref_dept
+    metadata = {"Series": series, "UUID": str(uuid.uuid4()), "fileId": str(uuid.uuid4()),
+                "description": description_to_use, "Filename": get_filename_from_cross_platform_path(file_path),
+                "FileReference": catalog_ref.removeprefix(f"{series}").strip().removeprefix("/").strip(),
```
techncl: Do you need to do this `f"{series}"` rather than just `series`?

python/ad-hoc-ingest/metadata_creator.py
```diff
@@ -65,7 +74,7 @@ def create_metadata_for_upload(row):
 
 def get_field_names():
```
techncl: Does this need to be a function? Could it not be a variable called "field_names"?

techncl (commented):
Looks good! Just some minor comments. Also, I noticed that there doesn't seem to be any positive path tests for the `main` function in ad_hoc_ingest; we don't actually know that main is working the way it's supposed to

python/ad-hoc-ingest/metadata_creator.py
```diff
@@ -4,34 +4,40 @@
 from pathlib import Path, PureWindowsPath, PurePosixPath
 
 import discovery_client
+from discovery_client import CollectionInfo
 
-def create_intermediate_metadata_dict(row, args):
+def create_intermediate_metadata_dict(description_override, row, args):
     catalog_ref = row["catRef"].strip()
```
techncl: This spelling of "catalogue" is the American one

sparkhi:
> Looks good! Just some minor comments. Also, I noticed that there doesn't seem to be any positive path tests for the `main` function in ad_hoc_ingest; we don't actually know that main is working the way it's supposed to

That's the main reason I made it small and moved stuff out into another function which is testable.

python/ad-hoc-ingest/metadata_creator.py
sparkhi: It does not make much difference as record details from discovery could be None so setting a default to empty string or None is not much different

python/ad-hoc-ingest/ad_hoc_ingest.py
sparkhi: 1 - Its fine. its not a question, its a flag

2 - the input is generated by archivists, I don't think its a problem

python/ad-hoc-ingest/metadata_creator.py
```diff
@@ -7,28 +7,28 @@
...
         description_to_use = row["description"].strip()
         discovery_iaid = ""
-    series = catalog_ref.split("/")[0].strip()
+    series = catalogue_ref.split("/")[0].strip()
     metadata = {"Series": series, "UUID": str(uuid.uuid4()), "fileId": str(uuid.uuid4()),
                 "description": description_to_use, "Filename": get_filename_from_cross_platform_path(file_path),
-                "FileReference": catalog_ref.removeprefix(f"{series}").strip().removeprefix("/").strip(),
+                "FileReference": catalogue_ref.removeprefix(series).strip().removeprefix("/").strip(),
                 "ClientSideOriginalFilepath": file_path, "IAID": discovery_iaid,
                 "formerRefDept": "" if former_ref_dept is None else former_ref_dept,
```
techncl, line 33: Now that you made `former_ref_dept` an empty string, you can just do `"formerRefDept": former_ref_dept` here

python/ad-hoc-ingest/metadata_creator.py
```diff
@@ -7,28 +7,28 @@
...
                 "formerRefTNA": "" if former_ref_tna is None else former_ref_tna}
```
techncl, line 34: Now that you made `former_ref_tna` an empty string, you can just do `"formerRefTNA": former_ref_tna` here

techncl (commented):
Look good! Just 2 minor things.

python/ad-hoc-ingest/metadata_creator.py
sparkhi: Sorry, that's what I meant in earlier response. The former ref from discovery can be 'None' so this line needs to check for 'None'. 
I cahnged the lines above to initialise to empty string just due to review comment. I much prefer 'None' in the initialisation when the value is unknown. "" is a valid value which can be a sensible default in certain cases but my preference is to not use valid values for unknowns.

techncl: Ohh, I get it now; that's my only major issue with Python: the lack of type information

techncl (dismissed):
👍

python/ad-hoc-ingest/argument_parser_builder_test.py
```diff
@@ -68,3 +69,21 @@ def test_should_throw_error_when_asset_source_is_used_but_no_value_given(self, c
...
+    def test_should_show_help_for_all_arguments_when_user_asks_for_help(self):
+        output = StringIO()
+
+        with redirect_stdout(output):
+            with self.assertRaises(SystemExit) as exit_code:
+                self.parser.parse_args(["-h"])
+
+        self.assertEqual(0, exit_code.exception.code)
+        help_message = output.getvalue()
+        print(help_message)
```
techncl: Is this supposed to be here?

techncl (commented):

techncl (approved):
👍

PR #1141, opened 2026-07-20
DR2-2916 Update DRI migration script

MancunianSam:
* Upload file object to DA object store bucket with the series as a tag, ifNoneMatch header set and the expected bucket owner
* Send the metadata location and files prefix in the sqs messages


techncl (approved):
👍

PR #1142, opened 2026-07-20
DR2 2917 optional files prefix for package builder

MancunianSam:
We’re assuming in the package builder that the files are under `{assetId}/` but for
things in the central bucket, they’ll be under `v1/{assetId}` We need an optional
filesPrefix parameter in the package builder message which will be used instead of {assetId}/if present.


scala/lambdas/utils/src/main/scala/uk/gov/nationalarchives/utils/ExternalUtils.scala
```diff
@@ -339,7 +339,7 @@ object ExternalUtils {
     case IngestStarted extends MessageStatus("Asset has started the ingest process.")
     case IngestError extends MessageStatus("There has been an error ingesting the asset.")
 
-  case class NotificationMessage(id: UUID, location: URI, messageId: Option[String] = None)
+  case class NotificationMessage(id: UUID, location: URI, filesPrefix: Option[String] = None, messageId: Option[String] = None)
```
techncl, line 342: I just noticed that we have a type alias for this called `LockTableMessage` (which we refer to sometimes) and we also have a case class in the court doc package builder called `LockTableMessage`, do we have to add the `filesPrefix` to that or is it only specific to TDR?

techncl (commented):
Looks good! Jus a minor comment. Also, would you mind updating the README?

scala/lambdas/utils/src/main/scala/uk/gov/nationalarchives/utils/ExternalUtils.scala
MancunianSam: This is only for the DRI migration so far and probably for adhoc and TDR ingests in the (very far) future. I don't think we'll ever put court documents through it.

techncl (approved):
👍

PR #1143, opened 2026-07-20
DR2-2918 Importer to use metadataLocation.

MancunianSam:
TDR are already sending this in the message from the export. There are other PRs to send this
for our other ingest routes.

We need this because the metadata file and the files themselves can be in different buckets now
so we need to stop guessing where the metadata file is.

This also takes an optional files prefix.


python/lambdas/preingest-importer/lambda_function_test.py
```diff
@@ -323,20 +341,6 @@ def test_should_raise_an_exception_when_the_file_does_not_exist_in_source_bucket
             "calling the ListObjects operation: Not Found'",
             str(ex.exception))
 
-    @patch('lambda_function.s3_client.list_objects')
-    def test_should_raise_exception_when_file_exists_but_corresponding_metadata_file_does_not_exist_in_source_bucket(
```
techncl, line 327: Shouldn't we modify this test rather than delete it since this error is thrown if there's no contents

python/lambdas/preingest-importer/lambda_function.py
```diff
@@ -23,24 +24,28 @@ def lambda_handler(event, context):
     for record in event["Records"]:
         body: dict[str, str] = json.loads(record["body"])
         asset_id = body["assetId"] if "assetId" in body else body["fileId"]
-        metadata_file_id = f"{asset_id}.metadata"
-        source_bucket = body["bucket"]
+        parsed_metadata_url = parse.urlparse(body["metadataLocation"])
+        metadata_file_id = parsed_metadata_url.path[1:]
+        metadata_source_bucket = parsed_metadata_url.netloc
+        files_source_bucket = body["bucket"]
+        files_prefix = body["filesPrefix"] if "filesPrefix" in body else asset_id
```
techncl: Could you not do `body.get("filesPrefix", asset_id)` instead?

techncl (commented):
Looks good! Just some minor comments. Also, would you mind updating the README?

techncl (approved):
👍

PR #1144, opened 2026-07-20
DR2-2919 Update other ingest routes to send metadata location.

MancunianSam:
The metadataLocation field is mandatory in the non-judgment ingest importer. This updates the
adhoc script to send it and the restore README to tell people to send it.


techncl (approved):
👍

PR #1146, opened 2026-07-20
DR2-2900 DRI migration changes

MancunianSam:
* Increase max session length. This will allow us to create larger DRI migration ingests.
* Allow importer to read from the object store
* Allow migration script to write to the object store.


techncl (approved):
👍

PR #1147, opened 2026-07-21
Add the slash for the prefix at the end.

MancunianSam:
This lets us find files under a "folder" in S3. Having it at the beginning doesn't work.

The DRI migration is sending filesPrefix without a trailing slash but I've added in some logic to check for it anyway.


techncl (approved):
👍

PR #1148, opened 2026-07-21
Migration fixes.

MancunianSam:
You need to call read on the file object and we need permission for the migration role to generate a data key.

The * in the permissions is ok because the key itself is restricted.

### DIFF
--- a/terraform/templates/iam_policy/dri_migration_policy.json.tpl
+++ b/terraform/templates/iam_policy/dri_migration_policy.json.tpl
@@ -30,7 +30,8 @@
     },
     {
       "Action": [
-        "kms:GenerateDataKey"
+        "kms:GenerateDataKey",
+	"kms:Decrypt"
       ],
       "Effect": "Allow",
       "Resource": "*",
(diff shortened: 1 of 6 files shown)


terraform/templates/iam_policy/dri_migration_policy.json.tpl
```diff
@@ -27,6 +27,18 @@
         "arn:aws:s3:::${object_store_bucket_name}",
         "arn:aws:s3:::${object_store_bucket_name}/*"
       ]
+    },
+    {
+      "Action": [
+        "kms:GenerateDataKey"
```
kurtismash: Add `kms:Decrypt`.

> To successfully make a PutObject request to encrypt an object with an AWS KMS key to Amazon S3, you need kms:GenerateDataKey permissions on the key. To download an object encrypted with an AWS KMS key, you need kms:Decrypt permissions for the key. To [perform a multipart upload](https://docs.aws.amazon.com/AmazonS3/latest/userguide/mpuoverview.html#mpuAndPermissions) to encrypt an object with an AWS KMS key, you must have the kms:GenerateDataKey and kms:Decrypt permissions for the key. ([link](https://docs.aws.amazon.com/AmazonS3/latest/userguide/UsingKMSEncryption.html))

kurtismash (changes requested):

kurtismash (approved):

PR #1149, opened 2026-07-21
DR2-2922 Stop listing objects in importer

MancunianSam:
In the non courtdoc importers, we’re currently doing list_bucket with a prefix of the `assetId` or `filesPrefix` and copying all of those objects across. The problem with this is that we have to trust that the source hasn’t copied files under that prefix without including them in the metadata file.

It also means that the lambda needs ListBucket permissions.

The more accurate way to do this is to construct the S3 keys using the ids from the metadata and copy those. So we read the metadata and construct a list of keys which are `{assetId or filesPrefix}/{fileId}`

If anything upstream isn’t following this convention then they should be.

I've removed the permissions from the importer side.

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -28,8 +28,8 @@ object Dependencies {
   lazy val fs2Core = "co.fs2" %% "fs2-core" % fs2Version
   lazy val fs2IO = "co.fs2" %% "fs2-io" % fs2Version
   lazy val fs2Reactive = "co.fs2" %% "fs2-reactive-streams" % fs2Version
-  lazy val jacksonCore = "tools.jackson.core" % "jackson-core" % "3.2.0"
-  lazy val jacksonDatabind = "tools.jackson.core" % "jackson-databind" % "3.1.4"
+  lazy val jacksonCore = "tools.jackson.core" % "jackson-core" % "3.2.1"
+  lazy val jacksonDatabind = "tools.jackson.core" % "jackson-databind" % "3.2.1"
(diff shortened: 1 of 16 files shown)


project/Dependencies.scala
techncl: I just realised that this version came out in 2021; the latest version is actually `0.14.16`

project/Dependencies.scala
```diff
@@ -28,8 +28,8 @@ object Dependencies {
   lazy val fs2Core = "co.fs2" %% "fs2-core" % fs2Version
   lazy val fs2IO = "co.fs2" %% "fs2-io" % fs2Version
   lazy val fs2Reactive = "co.fs2" %% "fs2-reactive-streams" % fs2Version
-  lazy val jacksonCore = "tools.jackson.core" % "jackson-core" % "3.2.0"
-  lazy val jacksonDatabind = "tools.jackson.core" % "jackson-databind" % "3.1.4"
+  lazy val jacksonCore = "tools.jackson.core" % "jackson-core" % "3.2.1"
+  lazy val jacksonDatabind = "tools.jackson.core" % "jackson-databind" % "3.2.1"
```
techncl: Should these two share a version?

techncl (commented):
Looks good! Would you mind updating the README?

project/Dependencies.scala
MancunianSam: That is odd. I've changed it to `0.14.6` and updated the tests that broke.

techncl (approved):
👍

PR #1151, opened 2026-07-22
DR2-2929 Don't copy tags or annotations

MancunianSam:
There are a load of permissions missing for copying objects with tags like those that come from the central bucket.
We don't need the tags for the DR2 ingest so they can be removed when we copy the files over.


techncl (approved):
👍

PR #1152, opened 2026-07-24
Dr2 2859 remove pa from the metric collector

techncl:


terraform/preingest.tf
```diff
@@ -33,7 +33,7 @@ module "dri_preingest" {
   ingest_lock_table_group_id_gsi_name          = local.ingest_lock_table_group_id_gsi_name
   ingest_raw_cache_bucket_name                 = local.ingest_raw_cache_bucket_name
   ingest_step_function_name                    = local.ingest_step_function_name
-  source_name                                  = "dri"
+  source_name                                  = lower(local.source_systems[index(local.source_systems, upper("dri"))])
```
MancunianSam: Do you need to do `upper("dri")`? Can you not just have `"DRI"` Same with the others.

python/lambdas/ingest-metric-collector/ingest_metric_collector.py
```diff
@@ -97,20 +97,21 @@ def get_flow_control_metrics(resources_prefix):
     return metric_data
 
 def lambda_handler(event, context):
+    source_systems = tuple(json.loads(os.environ["SOURCE_SYSTEMS"]))
```
MancunianSam, line 100: Do you need `tuple()` around this? It will be an array if you don't which should still work

MancunianSam (commented):
Looks good, thanks. Just a couple of questions.

python/lambdas/ingest-metric-collector/ingest_metric_collector.py
techncl: I did it just to ensure/indicate that it is immutable; should I remove it?

MancunianSam: It's fine if there's a reason, was just wondering.

MancunianSam (approved):

PR #1153, opened 2026-07-27
Remove extra arg for AnnotationDirective

MancunianSam:
We don't really need this at the moment and it's such a recent addition to boto3 that it's not included in the boto version that lambda has bundled

It's easier to remove it for now and add it back in later if necessary.


techncl (approved):
👍

PR #1155, opened 2026-07-28
DR2-2930 Stop assuming a role in the DRI migration script

MancunianSam:
We're going to be using a user instead. I've removed the get_clients method altogether and updated the tests.


techncl (approved):
👍

PR #1156, opened 2026-07-28
DR2-2932 Replace user with role in KMS key policy.

MancunianSam:
The reason I've done it like this with a hard coded string rather than with a data block is if I do
a data block then I have to add the role to the terraform policy which I don't really want to do for a temporary thing.


techncl (approved):
👍

PR #1158, opened 2026-07-28
DR2-2933 Remove PA enum

MancunianSam:
This is the last reference to the Parliament migration in code I can find.

I can't be 100% sure because PA is a very common string but I think I've got it all


techncl (approved):
👍

PR #1162, opened 2026-07-30
Parallel processing and S3 error handling.

MancunianSam:
This changes two things. First is to process assets in parallel 20 at a time.
The next is to handle If-None-Match PreconditionFailed errors from S3.
This means that the file has already been uploaded with that key. Without it, we can't
rerun a script which includes files already uploaded.
I'll accept the risk that a different file has been uploaded with that key as it's unlikely.

### DIFF
--- a/python/dri-migration/migrate/create_ingest_metadata.py
+++ b/python/dri-migration/migrate/create_ingest_metadata.py
@@ -194,7 +194,6 @@ def migrate(ic_db_path):
                 except ClientError as e:
                     error = e.response["Error"]
                     if error["Code"] == "PreconditionFailed" and error.get("Condition") == "If-None-Match":
-                        pass
                         print(f"Skipping asset {asset_id} as it already exists in {object_store_bucket}")
                     else:
                         raise e
--- a/python/dri-migration/migrate/create_ingest_metadata.py
+++ b/python/dri-migration/migrate/create_ingest_metadata.py
@@ -11,10 +11,12 @@ from collections import defaultdict
 from contextlib import closing
 from os import listdir
 from urllib import parse
+from concurrent.futures import ThreadPoolExecutor
 
 import boto3
 import oracledb
 from botocore.config import Config
+from botocore.exceptions import ClientError
 from pathlib import PureWindowsPath, PurePosixPath
 from dotenv import load_dotenv
 
@@ -45,7 +47,6 @@ config = Config(region_name="eu-west-2")
 sts_client = boto3.client("sts")
 s3_client = boto3.client("s3")
 sqs_client = boto3.client("sqs")
-
 
 def calculate_checksum(file_path: str, algorithm: str) -> str:
     try:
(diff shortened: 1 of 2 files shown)


python/dri-migration/migrate/create_ingest_metadata.py
```diff
@@ -159,51 +160,69 @@ def migrate(ic_db_path):
...
+                        Body=upload_file,
+                        Key=f"{prefix}/{asset_file_id}",
+                        Bucket=object_store_bucket,
+                        Tagging=tags,
+                        IfNoneMatch="*",
+                        ExpectedBucketOwner=object_store_account_number
+                    )
+                except ClientError as e:
+                    error = e.response["Error"]
+                    if error["Code"] == "PreconditionFailed" and error["Condition"] == "If-None-Match":
```
techncl: Is `Condition` always present in a `ClientError`? If not, would it be safer to do a `.get()`?

python/dri-migration/migrate/create_ingest_metadata.py
```diff
@@ -159,51 +160,69 @@ def migrate(ic_db_path):
...
+
+    all_sqs_messages = []
+    db_assets = []
+    grouped_asset_ids = list(grouped_assets.keys())
+    with ThreadPoolExecutor(max_workers=20) as executor:
+        count = 0
+        for migrated_assets, sqs_message in executor.map(migrate_asset, grouped_asset_ids):
+            count += 1
+            if count % 100 == 0:
+                print(f"Processed {count} assets")
```
techncl, line 219: Maybe f"Processed {count} grouped assets"?

python/dri-migration/migrate/create_ingest_metadata.py
```diff
@@ -159,51 +160,69 @@ def migrate(ic_db_path):
...
+                        Key=f"{prefix}/{asset_file_id}",
+                        Bucket=object_store_bucket,
+                        Tagging=tags,
+                        IfNoneMatch="*",
+                        ExpectedBucketOwner=object_store_account_number
+                    )
+                except ClientError as e:
+                    error = e.response["Error"]
+                    if error["Code"] == "PreconditionFailed" and error["Condition"] == "If-None-Match":
+                        pass
```
techncl: Should we maybe log this?

python/dri-migration/migrate/create_ingest_metadata.py
```diff
@@ -159,51 +160,69 @@ def migrate(ic_db_path):
 
     grouped_assets = group_assets(assets_with_redacted)
 
-    all_sqs_messages = []
-    assets = []
-
-    for asset_uuid, assets_list in grouped_assets.items():
+    def migrate_asset(asset_id):
```
techncl, line 163: Should this be `migrate_assets` since it's grouped assets?

techncl (commented):
Look good! Just some minor comments.

python/dri-migration/migrate/create_ingest_metadata.py
MancunianSam: I don't think grouped adds much. They're grouped in the code so assets with more than one file get grouped into a single asset but it's still a single asset and we're processing them one at a time.

python/dri-migration/migrate/create_ingest_metadata.py
MancunianSam: As above, it's still one asset, it just might have multiple files in it.

python/dri-migration/migrate/create_ingest_metadata.py
```diff
@@ -193,8 +193,9 @@ def migrate_asset(asset_id):
                     )
                 except ClientError as e:
                     error = e.response["Error"]
-                    if error["Code"] == "PreconditionFailed" and error["Condition"] == "If-None-Match":
+                    if error["Code"] == "PreconditionFailed" and error.get("Condition") == "If-None-Match":
                         pass
```
techncl: Should we remove the `pass`?

techncl (commented):
Just one minor thing

python/dri-migration/migrate/create_ingest_metadata.py
MancunianSam: Oh yeah, you're right, we don't need it, thanks

techncl (approved):
👍

PR #1163, opened 2026-07-30
Increase DRI aggregator timeout.

MancunianSam:
With around 3000 records, the lambda takes around 100 seconds which was timing out
which caused the aggregator to create a lot of smaller ingests.

This should be enough for a while.


techncl (approved):
👍

PR #1165, opened 2026-07-30
DR2-2943 Add metrics for intelligent caching

MancunianSam:


terraform/root_provider.tf
```diff
@@ -18,6 +18,10 @@ terraform {
     archive = {
       source = "hashicorp/archive"
     }
+    aws = {
+      source  = "hashicorp/aws"
+      version = "6.36.0"
```
techncl: There's a newer version of this `6.57.1`

techncl (commented):
Looks good! Tust a minor comment. Also, did we downgrade our `aws` version?

terraform/root_provider.tf
MancunianSam: This doesn't need to be here at all. I've removed it.

MancunianSam:
The lock file changes were related to the root_provider changes. Neither of them needed to be in here, I've reverted them both.

techncl (approved):
👍

PR #1167, opened 2026-07-31
DR2-2945 Eventbridge rules for unprocessed messages.

MancunianSam:
I've added this in for the CC queues including the two confirmer queues.

I don't think there's much point for the others as they're all feeding lambdas.


techncl (approved):
👍

PR #1168, opened 2026-07-31
Check AWS connectivity earlier in the process when dry run is false

sparkhi:


python/ad-hoc-ingest/ad_hoc_ingest_test.py
```diff
@@ -436,3 +427,75 @@ def test_should_report_error_when_discovery_is_not_reachable(self, mock_version_
...
+        mock_validate_aws_connectivity.assert_not_called()
+
+    @patch("ad_hoc_ingest.validate_aws_connectivity")
+    @patch("discovery_client.is_discovery_api_reachable")
+    @patch("ad_hoc_ingest.upload_files_to_ingest_bucket")
+    @patch("ad_hoc_ingest.get_input_dataset")
+    @patch("dataset_validator.validate_dataset")
+    @patch("ad_hoc_ingest.validate_arguments")
+    @patch("version_check.is_latest_version")
+    def test_main_method_should_not_call_aws_connectivity_validation_when_dry_run_mode_is_true(self, mock_version_check,  mock_validate_arguments, mock_validate_dataset, mock_get_input_dataset, mock_upload_to_ingest_bucket, mock_is_discovery_reachable, mock_validate_aws_connectivity):
```
MancunianSam: Should this test be called `...when_dry_run_mode_is_false`?

sparkhi: You are right, copy + paste and small change problem

python/ad-hoc-ingest/ad_hoc_ingest.py
```diff
@@ -149,6 +149,19 @@ def get_input_dataset(input_file_path):
         raise Exception("Unsupported input file format. Only CSV and Excel (xls, xlsx) files are supported for input")
     return data_set
 
+def validate_aws_connectivity():
+    try:
+        aws_interactions.get_account_number()
+        return
```
techncl: Is this `return` necessary?

python/ad-hoc-ingest/ad_hoc_ingest.py
```diff
@@ -149,6 +149,19 @@ def get_input_dataset(input_file_path):
...
+    except(ClientError, Exception) as e:
+        continue_without_connectivity = input(f"Warning: Unable to connect to AWS services because:\n{e}\nDo you want to continue? y/n")
+        if continue_without_connectivity.lower().strip() not in ["y", "yes"]:
+            mp.print_message("Exiting")
+            sys.exit(0)
+        else:
+            mp.print_message("Continuing without AWS connectivity")
+            return
```
techncl: Is this `return` necessary?

python/ad-hoc-ingest/ad_hoc_ingest.py
```diff
@@ -149,6 +149,19 @@ def get_input_dataset(input_file_path):
         raise Exception("Unsupported input file format. Only CSV and Excel (xls, xlsx) files are supported for input")
     return data_set
 
+def validate_aws_connectivity():
+    try:
+        aws_interactions.get_account_number()
+        return
+    except(ClientError, Exception) as e:
```
techncl, line 155: Shouldn't we be more specific with the error we're capturing?

python/ad-hoc-ingest/ad_hoc_ingest_test.py
```diff
@@ -436,3 +427,75 @@ def test_should_report_error_when_discovery_is_not_reachable(self, mock_version_
                     ad_hoc_ingest.main()
 
             self.assertEqual("Discovery API is not available for getting metadata information, terminating process", str(ex.exception))
+
+    @patch("aws_interactions.get_account_number")
+    def test_validate_aws_connectivity_should_return_true_when_aws_interaction_successfully_returns_an_account_number(self, mock_get_account_number):
```
techncl: But this is not returning "True". Maybe, "should not throw an error" or "should not exit"?

python/ad-hoc-ingest/ad_hoc_ingest_test.py
```diff
@@ -436,3 +427,75 @@ def test_should_report_error_when_discovery_is_not_reachable(self, mock_version_
...
+        mock_get_account_number.return_value = "67"
+        self.assertEqual(None, ad_hoc_ingest.validate_aws_connectivity())
+
+    @patch("aws_interactions.get_account_number")
+    @patch("builtins.input", return_value="y")
+    def test_validate_aws_connectivity_should_return_true_when_account_number_is_not_retrieved_but_user_wants_to_continue(self, mock_input, mock_get_account_number):
```
techncl: But this is not returning "True".

techncl (commented):

python/ad-hoc-ingest/ad_hoc_ingest.py
sparkhi: We are making use of the `get_account_number` just as a way to check AWS connection works, so we don't need to catch specific exception. Just an "it did not work" is what we need to know. The message from the underlying error is conveyed to the user

techncl (approved):
👍

PR #1198, opened 2026-08-07
Remove path-ignore and update python

techncl:


terraform/common.tf
```diff
@@ -107,13 +107,17 @@ locals {
     { id = "expire-object-delete-marker", status = "Enabled", expiration = { expired_object_delete_marker = true } }
   ]))
   source_systems = ["TDR", "COURTDOC", "ADHOC", "DRI", "DEFAULT"]
```
sparkhi, line 104: can we convert it to a map instead. This is feeling quite convoluted, the identification based on `index` is just moved from one place to other. instead if there is a map like
```
source_systems = {
    tdr = "TDR"
    courtdoc = "COURTDOC"
    ...
    ...
}
```

then the need to index goes away and further down it can just be used as `local.source_systems.tdr` or some such. and where you need all of them, it can change to use `values(local.source_systems)` or this can be another local variable if needed.

Having said that, we have a ticket to refactor common.tf and discuss it during the next team day, so all of this may get another treatment at that time

sparkhi (commented):
one small comment

terraform/common.tf
techncl: I thought about Maps when I first did this but Maps aren't ordered and since `source_systems` this array is now going to be the main source of truth (and is currently passed into the metric collector) , I thought that the order would matter; does it?

kurtismash: I'd go a lot further to make our ingest dynamic based on config rather than hardcoded Terraform. For example, a map of config objects as a var/local per env that is used throughout the Terraform. E.g.
```hcl
locals {
  source_systems = {
    COURTDOC = {
      ingest_flow_control_queue                      = false
      preingest_aggreator_max_batch_size             = 2000
      preingest_aggregator_secondary_batching_window = 300
    }
    DRI = {
      ingest_flow_control_queue                      = true
      ingest_flow_control_reserved_channels          = 0
      ingest_flow_control_probability                = 1
      preingest_aggreator_max_batch_size             = 2000
      preingest_aggregator_secondary_batching_window = 300
    }
    TDR = {
      ingest_flow_control_queue                      = false
      preingest_aggreator_max_batch_size             = 2000
      preingest_aggregator_secondary_batching_window = 300
    }
  }
}
```

techncl: Cool! That looks sensible. This PR was just a clean up one; should we add your suggestion to a ticket and then we can discuss it on a Backlog/Sprint planning meeting so that everyone can contribute is onboard?

kurtismash: Re-reading this thread, on the current question I don't think there's any benefit to the list or map. We could just have `local = "string"` to accomplish the same thing.

kurtismash: Yes @techncl, good idea!

techncl: I think there is a benefit of a list/map because it's easier to keep track of all the source systems and they could be easily deleted from, added to, moved around or passed into whatever project needs them without worrying that you've missed something

kurtismash: I think the concept is sound but the execution is duplicating effort. To add or remove a source system right now would require changing 2 lines, with only the string local being used elsewhere in the project. I think @sparkhi's suggestion of using a map is cleaner.

sparkhi (approved):

PR #1200, opened 2026-08-10
DR2-2960 Make flow control honour reserved channels

MancunianSam:
The `hasSpareChannels` value is used to determine whether or not to start a probability
based task after it has tried to start one on the reserved channels.

As it was, were doing `reservedChannelsCount < maxConcurrency` which for the prod config
is 1 < 6 which is always true so we were always starting a new task up to the maxConcurrency.

This now passes the running execution count in. So if maxConcurrency is 6, there is one reserved
channel and there are 5 existing executions, it won't start another one.

### DIFF
--- a/scala/lambdas/ingest-flow-control/src/main/scala/uk/gov/nationalarchives/ingestflowcontrol/Lambda.scala
+++ b/scala/lambdas/ingest-flow-control/src/main/scala/uk/gov/nationalarchives/ingestflowcontrol/Lambda.scala
@@ -27,25 +27,25 @@ class Lambda extends LambdaRunner[Option[Input], TaskOutput, Config, Dependencie
     val executionStarter = potentialInput.map(_.executionName).getOrElse("NO_EXECUTION_NAME")
 
     /** A recursive method to send task success to one and only one task per invocation. It takes a list of source systems, iterates over the list (recursively) to start a task on
-     * first system which has a reserved channel available
-     * @param sourceSystems
-     *   List of source systems to iterate over
-     * @param executionsBySystem
-     *   Map of SystemName -> count of running executions
-     * @param flowControlConfig
-     *   Flow control configuration
-     * @param taskExecutorName
-     *   Name of the task executor which sent the success
-     * @return
-     *   IO[String]: name of the task executor which sent success, a hard coded string "CONTINUE_TO_NEXT_SYSTEM" to indicate that no task was sent for the tried system or an empty
-     *   string to indicate that there is nothing more to do
-     */
+      * first system which has a reserved channel available
+      * @param sourceSystems
+      *   List of source systems to iterate over
+      * @param executionsBySystem
+      *   Map of SystemName -> count of running executions
+      * @param flowControlConfig
+      *   Flow control configuration
+      * @param taskExecutorName
+      *   Name of the task executor which sent the success
+      * @return
+      *   IO[String]: name of the task executor which sent success, a hard coded string "CONTINUE_TO_NEXT_SYSTEM" to indicate that no task was sent for the tried system or an empty
+      *   string to indicate that there is nothing more to do
+      */
     def startTaskOnReservedChannel(
-                                    sourceSystems: List[SourceSystem],
-                                    executionsBySystem: Map[String, Int],
-                                    flowControlConfig: FlowControlConfig,
-                                    taskExecutorName: String
-                                  ): IO[String] = {
+        sourceSystems: List[SourceSystem],
+        executionsBySystem: Map[String, Int],
+        flowControlConfig: FlowControlConfig,
+        taskExecutorName: String
+    ): IO[String] = {
       if sourceSystems.isEmpty then
         logInfo("Reserved channel: No more source systems to iterate over", taskExecutorName) >>
           IO.pure(taskExecutorName)
@@ -86,20 +86,20 @@ class Lambda extends LambdaRunner[Option[Input], TaskOutput, Config, Dependencie
     }
 
     /** A recursive method to send success to one of the systems, based on probability as configured in the Flow control configuration. It builds a probability ranges map based on
-     * the config (e.g. TDR -> (1, 25), FCL -> (25, 40) ... ). It then generates a random number between the minimum and maximum value over all probabilities and tries to schedule
-     * a task for that system. If there is no task waiting for the system, it recreates the probability map excluding that system from the config and generates a random number for
-     * the remaining systems only, thus making sure that the probabilities are honoured over all iterations.
-     *
-     * The probability of each system is kept intact in relation to each other, even if one of the systems does not have a waiting task. e.g. if there are 3 systems, one, two, and
-     * three with probabilities of 25, 55, 20, the ranges being 1-26, 26-81, 81-101. For this example, let's assume that system "two" does not have a waiting task. Iteration 1 -
-     * random number generated is 30 (which corresponds to system "two"), since system "two" does not have a waiting task, Iteration 2 - a new ranges map is constructed by
-     * excluding system "two" (1-26, 26-47) and a new number is generated between 1 and 47 this ensures that the probability in relation to each other is kept intact for the
-     * remaining systems
-     * @param sourceSystems
-     *   List of source systems to iterate for starting a task
-     * @return
-     *   IO[String]: name of the task executor which sent success, a hard coded string "CONTINUE_TO_NEXT_SYSTEM" to indicate that no task was sent for the tried system
-     */
+      * the config (e.g. TDR -> (1, 25), FCL -> (25, 40) ... ). It then generates a random number between the minimum and maximum value over all probabilities and tries to schedule
+      * a task for that system. If there is no task waiting for the system, it recreates the probability map excluding that system from the config and generates a random number for
+      * the remaining systems only, thus making sure that the probabilities are honoured over all iterations.
+      *
+      * The probability of each system is kept intact in relation to each other, even if one of the systems does not have a waiting task. e.g. if there are 3 systems, one, two, and
+      * three with probabilities of 25, 55, 20, the ranges being 1-26, 26-81, 81-101. For this example, let's assume that system "two" does not have a waiting task. Iteration 1 -
+      * random number generated is 30 (which corresponds to system "two"), since system "two" does not have a waiting task, Iteration 2 - a new ranges map is constructed by
+      * excluding system "two" (1-26, 26-47) and a new number is generated between 1 and 47 this ensures that the probability in relation to each other is kept intact for the
+      * remaining systems
+      * @param sourceSystems
+      *   List of source systems to iterate for starting a task
+      * @return
+      *   IO[String]: name of the task executor which sent success, a hard coded string "CONTINUE_TO_NEXT_SYSTEM" to indicate that no task was sent for the tried system
+      */
     def startTaskBasedOnProbability(sourceSystems: List[SourceSystem]): IO[String] = {
       sourceSystems match
         case Nil =>
@@ -252,13 +252,13 @@ class Lambda extends LambdaRunner[Option[Input], TaskOutput, Config, Dependencie
   }
 
   private def logInfo(
-                       message: String,
-                       executionName: String,
-                       currentSystem: String = "",
-                       remainingSystems: String = "",
-                       queuedTimeAndExecution: String = "",
-                       resumedExecution: String = ""
-                     ): IO[Unit] = {
+      message: String,
+      executionName: String,
+      currentSystem: String = "",
+      remainingSystems: String = "",
+      queuedTimeAndExecution: String = "",
+      resumedExecution: String = ""
+  ): IO[Unit] = {
     logger.info(
       Map(
         "executionName" -> executionName,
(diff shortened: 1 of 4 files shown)


scala/lambdas/ingest-flow-control/src/test/scala/uk/gov/nationalarchives/ingestflowcontrol/LambdaTest.scala
```diff
@@ -131,6 +131,39 @@ class LambdaTest extends AnyFlatSpec with EitherValues with TableDrivenPropertyC
     lambdaRunResult.finalStepFnExecutions.find(_.taskToken == "a-task-already-running").exists(_.taskTokenSuccess) should be(false)
   }
 
+  "lambda" should s"add new task to dynamo but not send success when there are is a max concurrency of 6, 5 running tasks and 1 reserved channel" in {
```
techncl: 1. you have "...there are is" in the test name
2. Maybe this test name could be more generic, e.g. "...when running tasks + reserved channels is less than max concurrency" as the test name makes it seems as though it will only add a new task when that specific config is in place?

scala/lambdas/ingest-flow-control/src/test/scala/uk/gov/nationalarchives/ingestflowcontrol/LambdaTest.scala
```diff
@@ -131,6 +131,39 @@ class LambdaTest extends AnyFlatSpec with EitherValues with TableDrivenPropertyC
...
+      StepFunctionExecution("DRI", s"DRI_$i")
+    }.toList
+    val input = Option(Input("DRI_execution_name_2", "task-token-for-dri"))
+
+    val lambdaRunResult = runLambda(input, Nil, initialConfig, existingExecutions, predictableRandomNumberSelector())
+
+    lambdaRunResult.result.isRight should be(true)
+    lambdaRunResult.finalItemsInTable should have length 1
+    lambdaRunResult.finalItemsInTable.map(_.taskToken).contains("task-token-for-dri") should be(true)
+  }
```
techncl, line 152: Is it worth adding ` lambdaRunResult.finalStepFnExecutions.forall(_.taskTokenSuccess) should be(true)`?

scala/lambdas/ingest-flow-control/src/test/scala/uk/gov/nationalarchives/ingestflowcontrol/LambdaTest.scala
```diff
@@ -131,6 +131,39 @@ class LambdaTest extends AnyFlatSpec with EitherValues with TableDrivenPropertyC
...
+
+  "lambda" should s"not send task success to a running task when there are is a max concurrency of 6, 5 running tasks and 1 reserved channel" in {
```
techncl: Maybe this test name could be more generic, e.g. "...when running tasks + reserved channels is less than max concurrency" as the test name makes it seems as though it will only add a new task when that specific config is in place?

scala/lambdas/ingest-flow-control/src/test/scala/uk/gov/nationalarchives/ingestflowcontrol/LambdaTest.scala
```diff
@@ -431,13 +464,13 @@ class LambdaTest extends AnyFlatSpec with EitherValues with TableDrivenPropertyC
       ),
       true
     )
-    configWithSpareChannels.hasSpareChannels should be(true)
+    configWithSpareChannels.hasSpareChannels(0) should be(true)
   }
 
   "FlowControlConfig" should "indicate lack of spare channels when reserved channels equal the maximum concurrency" in {
```
techncl: Should this be "...when reserved channels + running executions"

scala/lambdas/ingest-flow-control/src/test/scala/uk/gov/nationalarchives/ingestflowcontrol/LambdaTest.scala
```diff
@@ -431,13 +464,13 @@ class LambdaTest extends AnyFlatSpec with EitherValues with TableDrivenPropertyC
...
     val configWithAllChannelsReserved =
       Lambda.FlowControlConfig(4, List(Lambda.SourceSystem("SystemOne", 1, 25), Lambda.SourceSystem("SystemTwo", 1, 35), Lambda.SourceSystem("DEFAULT", 2, 40)), true)
-    configWithAllChannelsReserved.hasSpareChannels should be(false)
+    configWithAllChannelsReserved.hasSpareChannels(4) should be(false)
```
techncl: Why 4 and not 0? I was thinking that "0" would indicate that even though there are 0 running executions, there are still no spare channels as they are all reserved

scala/lambdas/ingest-flow-control/src/main/scala/uk/gov/nationalarchives/ingestflowcontrol/Lambda.scala
```diff
@@ -224,6 +224,7 @@ class Lambda extends LambdaRunner[Option[Input], TaskOutput, Config, Dependencie
       flowControlConfig <- dependencies.ssmClient.getParameter[FlowControlConfig](config.configParamName)
       _ <- writeTaskToQueueTable(flowControlConfig)
       runningExecutions <- dependencies.stepFunctionClient.listStepFunctions(config.stepFunctionArn, Running)
+      _ <- logInfo(s"Found ${runningExecutions.length} running executions", executionStarter)
```
techncl, line 228: Maybe "Currently" instead of "Found"?

scala/lambdas/ingest-flow-control/src/main/scala/uk/gov/nationalarchives/ingestflowcontrol/Lambda.scala
```diff
@@ -336,7 +337,7 @@ object Lambda {
     require(sourceSystems.map(_.systemName).contains(default), "Missing 'DEFAULT' system in the configuration")
 
     val hasReservedChannels: Boolean = reservedChannelsCount > 0
-    val hasSpareChannels: Boolean = reservedChannelsCount < maxConcurrency
+    def hasSpareChannels(runningCount: Int): Boolean = runningCount < (maxConcurrency - reservedChannelsCount)
```
techncl: Is it worth saving (maxConcurrency - reservedChannelsCount) to a variable instead?

techncl (commented):
Looks good! Some minor comments. Also, would you mind updating the README with this please?

scala/lambdas/ingest-flow-control/src/test/scala/uk/gov/nationalarchives/ingestflowcontrol/LambdaTest.scala
MancunianSam: 1. I fixed this one, copilot caught it.
2. Yeah ok

scala/lambdas/ingest-flow-control/src/test/scala/uk/gov/nationalarchives/ingestflowcontrol/LambdaTest.scala
MancunianSam: I've merged these two tests now so it does have a line checking the executions. It should be `lambdaRunResult.finalStepFnExecutions.forall(!_.taskTokenSuccess) should be(true)` though as none of them should have sendTaskSuccess sent.

scala/lambdas/ingest-flow-control/src/test/scala/uk/gov/nationalarchives/ingestflowcontrol/LambdaTest.scala
MancunianSam: This has changed now. The original logic wasn't right.

scala/lambdas/ingest-flow-control/src/main/scala/uk/gov/nationalarchives/ingestflowcontrol/Lambda.scala
MancunianSam: Found is fairly standard, I'll leave it as is.

scala/lambdas/ingest-flow-control/src/main/scala/uk/gov/nationalarchives/ingestflowcontrol/Lambda.scala
MancunianSam: This has changed as well, the logic is a bit different.

scala/lambdas/ingest-flow-control/src/test/scala/uk/gov/nationalarchives/ingestflowcontrol/LambdaTest.scala
```diff
@@ -431,13 +470,13 @@ class LambdaTest extends AnyFlatSpec with EitherValues with TableDrivenPropertyC
...
       true
     )
-    configWithSpareChannels.hasSpareChannels should be(true)
+    configWithSpareChannels.hasSpareChannels(Map("SystemOne" -> 1, "DEFAULT" -> 1)) should be(true)
   }
 
-  "FlowControlConfig" should "indicate lack of spare channels when reserved channels equal the maximum concurrency" in {
+  "FlowControlConfig" should "indicate lack of spare channels when reserved channels + running executions equal the maximum concurrency" in {
     val configWithAllChannelsReserved =
       Lambda.FlowControlConfig(4, List(Lambda.SourceSystem("SystemOne", 1, 25), Lambda.SourceSystem("SystemTwo", 1, 35), Lambda.SourceSystem("DEFAULT", 2, 40)), true)
```
sparkhi: I think this test config should change to make justice to the test name. at the moment the config has 4 max concurrency and total 4 reserved channels. I think, just making the maxConcurrency to 7 instead of 4 would suffice. Maybe change the variable name as well to reflect it

techncl:
Would you mind updating the README?

scala/lambdas/ingest-flow-control/src/test/scala/uk/gov/nationalarchives/ingestflowcontrol/Helpers.scala
```diff
@@ -18,7 +18,7 @@ object Helpers {
 
   def notImplemented[T]: IO[Nothing] = IO.raiseError(new Exception("Not implemented"))
 
-  def predictableRandomNumberSelector(selected: Int = 10): (Int, Int) => Int = (min, max) => if selected > max then max else selected
+  def predictableRandomNumberSelector(selected: Int = 10): (Int, Int) => Int = (min, max) => if selected > max then max - 1 else selected
```
sparkhi, line 21: I think this was okay to not need -1 
It was there to just protect from the default value used in tests where a selection range could be 1 to 6 but the default "selected" defined here is 10, in such case, it will return 6 as the selection.
The range case class used in the implementation treats the lower value as included i nthe range and upper value as excluded

sparkhi (commented):
Just two small comments

scala/lambdas/ingest-flow-control/src/main/scala/uk/gov/nationalarchives/ingestflowcontrol/Lambda.scala
```diff
@@ -336,7 +338,12 @@ object Lambda {
     require(sourceSystems.map(_.systemName).contains(default), "Missing 'DEFAULT' system in the configuration")
 
     val hasReservedChannels: Boolean = reservedChannelsCount > 0
-    val hasSpareChannels: Boolean = reservedChannelsCount < maxConcurrency
+    def hasSpareChannels(executionsMap: Map[String, Int]): Boolean =
+      val totalReservedChannels = sourceSystems.foldLeft(0) { (totalReserved, sourceSystem) =>
+        totalReserved + max(sourceSystem.reservedChannels - executionsMap.getOrElse(sourceSystem.systemName, 0), 0)
```
techncl, line 343: Why does this need to be wrapped in `max`? Is it in case a source system is not running on a reserve channel and therefore could be a minus number?

scala/lambdas/ingest-flow-control/src/test/scala/uk/gov/nationalarchives/ingestflowcontrol/Helpers.scala
MancunianSam: I put this in here because I had a test that simulated the flow control we've got now in prod with the probability of DRI set to 1 and it was failing without this -1. And the tests still pass with it set like that so I think this is fine to leave.

scala/lambdas/ingest-flow-control/src/test/scala/uk/gov/nationalarchives/ingestflowcontrol/LambdaTest.scala
MancunianSam: Good point. I've changed it to 8 concurrency and this passes because 4 reserved + 4 running. I've changed the variable name as well.

scala/lambdas/ingest-flow-control/src/main/scala/uk/gov/nationalarchives/ingestflowcontrol/Lambda.scala
MancunianSam: Yep.

techncl (approved):
👍

PR #1201, opened 2026-08-10
DR2-2974 Revert courtdoc aggregator changes

MancunianSam:
As part of the flow control config in DR2-2909 we changed the batching window for the courtdoc preingest to 4 hours to try and aggregate lots of court docs into one. This isn’t actually working and it’s causing the e2e tests to fail because the courtdoc preingest waits for 4 hours.

This needs to be reverted and the flow control in prod to be updated to remove TDR and add a reserved channel for DEFAULT. This means that TDR and COURTDOC can share the reserved channel. As neither of these ingests usually take very long, this should be fine.

### DIFF
--- a/terraform/common.tf
+++ b/terraform/common.tf
@@ -124,19 +124,14 @@ locals {
       enabled        = true,
       sourceSystems = [
         {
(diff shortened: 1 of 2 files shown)


techncl (approved):
👍

PR #1203, opened 2026-08-10
Dr2 2925 update the courtdoc importer code

sparkhi:


scala/lambdas/preingest-courtdoc-importer/src/main/scala/uk/gov/nationalarchives/preingestcourtdocimporter/Lambda.scala
```diff
@@ -21,115 +18,59 @@ import scala.jdk.CollectionConverters.*
...
-                        tarEntry.getName -> id
-                      }
-                  }
-                )
-              } else Stream.empty
-            } ++
-            unarchiveAndUploadToS3(tarInputStream)
-        }
+    val metadataFolder = "/out"
+    val dataFileFolder = "/out/data"
```
techncl: Should we make this s"${metadataFolder}/data"?

scala/lambdas/preingest-courtdoc-importer/src/main/scala/uk/gov/nationalarchives/preingestcourtdocimporter/Lambda.scala
```diff
@@ -21,115 +18,59 @@ import scala.jdk.CollectionConverters.*
...
+          log(Map("error" -> err.getMessage, "s3FolderName" -> treInput.parameters.s3FolderName))(err.getMessage)
         }
-        fileId <- IO.fromOption(fileNameToFileIds.get(s"$batchRef/${treMetadata.parameters.TRE.payload.filename}"))(
-          new RuntimeException(s"Cannot find file name for file belonging to $batchRef")
-        )
-        irrelevantFilesFromTre = fileNameToFileIds.values.filter(!List(fileId, metadataFileId).contains(_))
-        _ <- IO.whenA(irrelevantFilesFromTre.nonEmpty) {
-          dependencies.s3.deleteObjects(outputBucket, irrelevantFilesFromTre.toList.map(_.toString)) >>
-            log(Map("fileReference" -> treMetadata.parameters.TDR.`File-Reference`.orNull))("Deleted unused TRE objects from the root of S3")
+        originalFileKey <- IO.pure(s"${treInput.parameters.s3FolderName}$dataFileFolder/${treMetadata.parameters.TRE.payload.filename}")
```
techncl: 1. Maybe `treFileKey` instead of `originalFileKey` as "original" seems like it's going to be replaced; we're just copying from one bucket to another? Or `fileSourceKey` to keep it consistent with the `metadataSourceKey`?
2. This doesn't need to be wrapped in pure, it can just be
```
originalFileKey = s"${treInput.parameters.s3FolderName}$dataFileFolder/${treMetadata.parameters.TRE.payload.filename}"

File: scala/lambdas/preingest-courtdoc-importer/src/main/scala/uk/gov/nationalarchives/preingestcourtdocimporter/Lambda.scala [hunk 3]
```diff
@@ -21,115 +18,59 @@ import scala.jdk.CollectionConverters.*
...
+        _ <- dependencies.s3.copy(treInput.parameters.s3Bucket, metadataSourceKey, outputBucket, metadataDestinationKey).onError { err =>
+          log(Map("error" -> err.getMessage, "metadataFileId" -> metadataSourceKey))(err.getMessage)
         }
+
         tdrId = treMetadata.parameters.TDR.`UUID`
         _ <- dependencies.sqsClient.sendMessage(config.outputQueueUrl)(
-          Message(tdrId, fileId, s"s3://${config.outputBucket}/$metadataFileId", treInput.properties.flatMap(_.messageId), treInput.parameters.skipSeriesLookup)
+          Message(tdrId, dr2FileKey, s"s3://${config.outputBucket}/$dr2FileKey", treInput.parameters.skipSeriesLookup)
         )
+        _ <- log(Map("batchRef" -> batchRef))(s"Finished processing batch $batchRef and imported file $dr2FileKey")
```
techncl: Should we add `metadataDestinationKey` too?

scala/lambdas/preingest-courtdoc-importer/src/main/scala/uk/gov/nationalarchives/preingestcourtdocimporter/Lambda.scala
```diff
@@ -21,115 +18,59 @@ import scala.jdk.CollectionConverters.*
...
           .compile
-          .toList
-        parsedJson <- IO.fromOption(contentString.headOption)(
-          new RuntimeException(
-            "Error parsing metadata.json.\nPlease check that the JSON is valid and that all required fields are present"
-          )
+          .string
+        parsedJson <- IO.fromEither(
+          decode[TREMetadata](contentJson).leftMap { error =>
+            new RuntimeException("Error parsing metadata.json.\nPlease check that the JSON is valid and that all required fields are present", error)
```
MancunianSam: I'm not sure we need this custom exception. If this happens it will be caught and logged on line 57 and that line only logs the message and this message is generic without any information as to what's wrong. 
If you just do 
```scala
parsedJson <- IO.fromEither(decode[TREMetadata](contentJson))
```
that should give more information.

scala/lambdas/preingest-courtdoc-importer/src/main/scala/uk/gov/nationalarchives/preingestcourtdocimporter/Lambda.scala
MancunianSam: You don't really need the IO wrapper here, you can just do 
```scala
originalFileKey = s"${treInput.parameters.s3FolderName}$dataFileFolder/${treMetadata.parameters.TRE.payload.filename}"
```

scala/lambdas/preingest-courtdoc-importer/src/main/scala/uk/gov/nationalarchives/preingestcourtdocimporter/Lambda.scala
```diff
@@ -21,115 +18,59 @@ import scala.jdk.CollectionConverters.*
...
-        _ <- IO.whenA(irrelevantFilesFromTre.nonEmpty) {
-          dependencies.s3.deleteObjects(outputBucket, irrelevantFilesFromTre.toList.map(_.toString)) >>
-            log(Map("fileReference" -> treMetadata.parameters.TDR.`File-Reference`.orNull))("Deleted unused TRE objects from the root of S3")
+        originalFileKey <- IO.pure(s"${treInput.parameters.s3FolderName}$dataFileFolder/${treMetadata.parameters.TRE.payload.filename}")
+        dr2FileKey = dependencies.uuidGenerator()
+
+        _ <- dependencies.s3.copy(treInput.parameters.s3Bucket, originalFileKey, outputBucket, dr2FileKey.toString).onError { err =>
+          log(Map("error" -> err.getMessage, "fileId" -> dr2FileKey.toString))(err.getMessage)
+        }
+        _ <- dependencies.s3.copy(treInput.parameters.s3Bucket, metadataSourceKey, outputBucket, metadataDestinationKey).onError { err =>
```
MancunianSam, line 61: This is copying the metadata file to the raw cache bucket with its `s"TRE-$batchRef-metadata.json"` key but the old importer just used a random UUID so we should keep that. Something like 
```
metadataSourceKey = dependencies.uuidGenerator().toString
        _ <- dependencies.s3.copy(treInput.parameters.s3Bucket, metadataSourceKey, outputBucket, metadataSourceKey)
```

scala/lambdas/preingest-courtdoc-importer/src/test/scala/uk/gov/nationalarchives/preingestcourtdocimporter/TestUtils.scala
```diff
@@ -55,25 +52,60 @@ object TestUtils:
...
-        val content = Flux
-          .from(publisher)
-          .toIterable
-          .asScala
-          .toList
-          .head
-        S3Object(bucket, key, content) :: existing
-      }
-      .map(_ => CompletedUpload.builder.response(PutObjectResponse.builder.build).build)
+    override def upload(
```
MancunianSam, line 96: This method isn't called in the code any more so you can get rid of this and return `IO.stub`
You are using it in the tests as part of the copy method but you can simplify that, something like this, I haven't tested this thoroughly.

```scala
      errors.raise(_.copy, "Error copying files") >> ref.update { objects =>
        objects.map { obj =>
          if obj.bucket == sourceBucket && obj.key == sourceKey then
            obj.copy(bucket = destinationBucket, key = destinationKey)
          else obj
        }
      }.map { _ =>
        CompletedCopy
          .builder()
          .response(CopyObjectResponse.builder().build())
          .build()
      }
```

scala/lambdas/preingest-courtdoc-importer/src/test/scala/uk/gov/nationalarchives/preingestcourtdocimporter/TestUtils.scala
```diff
@@ -103,58 +134,48 @@ object TestUtils:
...
-      val entry = new TarArchiveEntry(s"$batchReference/$fileName")
-      entry.setSize(content.length)
-      tarOut.putArchiveEntry(entry)
-      tarOut.write(content.getBytes)
-      tarOut.closeArchiveEntry()
+  def initialTestDataInTRECommonBucket(metadata: TREMetadata, batchReference: String, s3FolderName: String): Map[String, ByteBuffer] =
+    initialTestDataInTRECommonBucket(metadata.asJson.noSpaces, batchReference, s3FolderName)
+
+  def initialTestDataInTRECommonBucket(content: String, batchReference: String, s3FolderName: String): Map[String, ByteBuffer] = {
+    val files = Map(s"$s3FolderName/out/data/Test.docx" -> Array.fill(100)("a").mkString, "unused.txt" -> "", s"$s3FolderName/out/TRE-$batchReference-metadata.json" -> content)
```
MancunianSam, line 163: There's an unused.txt file here. Is it worth adding a test to make sure that doesn't get copied over?

scala/lambdas/preingest-courtdoc-importer/src/main/scala/uk/gov/nationalarchives/preingestcourtdocimporter/Lambda.scala
```diff
@@ -162,8 +112,8 @@ object Lambda:
 
   case class Config(outputBucket: String, outputQueueUrl: String) derives ConfigReader
 
-  case class TREInputProperties(messageId: Option[String])
+  case class TREInputProperties(messageType: String, `function`: String, producer: String, executionId: String, parentExecutionId: String, timestamp: String)
```
MancunianSam: I don't think we're using any of these now that we're not bothering with the message id so I think you can just get rid of this case class altogether.

scala/lambdas/preingest-courtdoc-importer/src/main/scala/uk/gov/nationalarchives/preingestcourtdocimporter/Lambda.scala
```diff
@@ -142,18 +83,27 @@ object Lambda:
   private val chunkSize: Int = 1024 * 64
```
MancunianSam: This doesn't seem to be used, you can remove it.

scala/lambdas/preingest-courtdoc-importer/src/test/scala/uk/gov/nationalarchives/preingestcourtdocimporter/LambdaTest.scala
```diff
@@ -47,85 +139,72 @@ class LambdaTest extends AnyFlatSpec with EitherValues {
...
-
-    val (res, _, _) = runLambda(List(S3Object("inputBucket", fileName, content)), event(), Option(Errors(download = true)))
+    val (res, _, _) = runLambda(s3Objects, event(), Option(Errors(download = true)))
 
     res.isLeft should equal(true)
     res.left.value.getMessage should equal("Error downloading files")
   }
 
-  "lambda handler" should "error if there is an error uploading the extracted file" in {
+  "lambda handler" should "error if there is an error uploading the file" in {
```
MancunianSam: You're not uploading any more, just copying so the test name and any other references to upload needs to change to copy.

MancunianSam (commented):
Looks good, thanks. Just a few bits to check. Can you update the README as well please.

scala/lambdas/preingest-courtdoc-importer/src/test/scala/uk/gov/nationalarchives/preingestcourtdocimporter/LambdaTest.scala
techncl: Should we also check that the keys are the expected keys? Same question for the other tests

scala/lambdas/preingest-courtdoc-importer/src/main/scala/uk/gov/nationalarchives/preingestcourtdocimporter/Lambda.scala
sparkhi: Aah good. I had overlooked. I thought, only the data file was copied with UUID, and metadata json was copied as it is. That also explains why I ended up with just one UUID in the generator in test. I'll reinstate the iterator in the tests as well with one more uuid. It also means, I was not sending the correct destination in the message.

techncl (commented):
Looks good! Just some minor comments. Also:
1. Is it worth adding 1 or more "message decoder" tests in order to confirm a message is thrown if a field in "properties" and/or "parameters" is missing?
2. Could you update the README, please?

scala/lambdas/preingest-courtdoc-importer/src/test/scala/uk/gov/nationalarchives/preingestcourtdocimporter/TestUtils.scala
sparkhi: In the "should copy only data and metadata file", we already have the check for count to be 2 and I'll add the key checking.

MancunianSam (dismissed):

MancunianSam:
The tests are failing because of a vulnerability in one of the dependencies. It's fixed here https://github.com/nationalarchives/dr2-ingest/pull/1209 so if you wait for that to be merged it should be fine.

terraform/preingest.tf
```diff
@@ -103,8 +103,8 @@ module "court_document_preingest" {
     filter_policy = templatefile("${path.module}/templates/sns/tre_live_stream_filter_policy.json.tpl", {})
   } : null
   source_name                = "courtdoc"
-  bucket_kms_arn             = module.tre_config.terraform_config["prod_s3_court_document_pack_out_kms_arn"]
-  copy_source_bucket_arn     = module.tre_config.terraform_config[local.tre_environment_name]["s3_court_document_pack_out_arn"]
+  bucket_kms_arn             = module.tre_config.terraform_config["prod_s3_common_kms_arn"]
```
MancunianSam: I think this needs to be 
```hcl
module.tre_config.terraform_config["${local.tre_environment_name}_s3_common_kms_arn"]
```
so we get the KMS key for the right environmnet.

MancunianSam (dismissed):

MancunianSam (approved):

PR #1204, opened 2026-08-11
DR2-2979 Trigger flow control in e2e tests

MancunianSam:
I tried to do this within the JUnit/Cucumber tests but they're not really set up for that so there's now a bash script which will run invoke-function every 5 minutes.


techncl (approved):
👍

PR #1206, opened 2026-08-12
DR2-2847 Parallelise test runs.

MancunianSam:
This uses the new `parallel` step for most and then a normal jobs matrix
for the sbt tests.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -8,33 +8,72 @@ permissions:
   id-token: write
   contents: write
 jobs:
-  test:
+  dependency_review:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
       - uses: sbt/setup-sbt@bfea3c5f48abd221b04a6df4798aa5eb8b6a2baf # v1.5.6
       - uses: scalacenter/sbt-dependency-submission@d84eef4c09e633bcf5f113bcad7fd5e9af1baee9 # v3.2.3
       - uses: actions/dependency-review-action@a1d282b36b6f3519aa1f3fc636f609c47dddb294 # v5.0.0
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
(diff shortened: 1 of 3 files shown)


.github/workflows/test.yml
```diff
@@ -8,33 +8,72 @@ permissions:
...
+        - name: Run Python tests
+          run: |
+            find . -name requirements.txt | xargs -I {} pip install -r {}
+            pytest
+          env:
+            AWS_DEFAULT_REGION: eu-west-2
+            AWS_ACCESS_KEY_ID: test
+            AWS_SECRET_ACCESS_KEY: test
+        - name: Run Kotlin tests
+          run: (cd kotlin/e2e-tests/lib && gradle test)
```
techncl, line 54: I just realised that we don't have a Kotlin formatter; should we?

.github/workflows/test.yml
```diff
@@ -8,33 +8,72 @@ permissions:
...
+  discover_sbt_tests:
```
techncl, line 55: Maybe `get_project_ids` or `get_sbt_project_ids`?

scala/lambdas/build.sbt
```diff
@@ -4,6 +4,25 @@ import sbtassembly.Log4j2MergeStrategy
...
+    structure.allProjectRefs
+      .map(_.project)
+      .distinct
+      .filterNot(List("ingestLambdasRoot", "root", "utils").contains)
+      .sorted
+
+  val json =
+    projectIds.map(id => "\"" + id + "\"").mkString("[", ",", "]")
+
+  println(json)
```
techncl, line 22: Is this deliberate?

techncl (commented):

.github/workflows/test.yml
MancunianSam: Not now, maybe if we're really bored.

.github/workflows/test.yml
MancunianSam: They're not projects, they're modules, and we're not getting ids, we're getting the names. This is ok as is, it's reasonably clear what it's doing.

scala/lambdas/build.sbt
MancunianSam: Yes, that's the thing that outputs what we need to console. It won't work without it.

techncl (approved):
👍

PR #1208, opened 2026-08-13
DR2-2944 Add more logging and move dependencies

MancunianSam:
Because IO doesn't memoise whatever is inside it, the constructor for the
s3 and dynamo client are being called every time the lamdba is run which is why
it's running out of file handles.

Constructing it outside the handler means that it is only created once every time the lambda
starts from cold.

I've also added in a small amount of extra logging and added the log4j file that allows the logging
to work


techncl (approved):
👍

PR #1211, opened 2026-08-19
NOJIRA Re-enable non-prod environments

kurtismash:
Re-enables Flow Control in non-prod environments now that the large ingest has completed. When this deploys it'll revert  the manual configuration changes we made to get the large ingest through.


techncl (approved):
👍

PR #1213, opened 2026-08-21
Dr2 3000 pass values into flow control then queue table

techncl:


terraform/README.md
```diff
@@ -119,6 +119,10 @@ Example: preingest_court_document_step_function_arn        = module.court_docume
 
 ### Troubleshooting:
 
+1. if after running `terraform init`, you get
+`Error refreshing state: Unable to access object "terraform.state" in S3 bucket "{management_bucket}" operation error S3: HeadObject, https response error StatusCode: 403...`
+you need to `aws sso login...` login with your aws credentials for the management account
```
MancunianSam: This isn't strictly true. You need to assume the role before `init` will work from a clean install so if you have permissions errors, you'll need to unset any access keys, then log in, then re-assume the role again. 

While I'm thinking about it, the instructions further up say to do `terraform init` before assuming the role and that won't work either.

scala/lambdas/ingest-flow-control/README.md
```diff
@@ -43,47 +47,77 @@ In the configuration shown above,
...
-   5. If there are not enough reserved channels for the system, it carries on iterating over the remaining systems to try and call "sendTaskSuccess" on any running execution
-   6. It then calculates whether there are spare channels in the system. It adds the number of running executions to the number of reserved channels, excluding those reserved channels which are already in use. If this is less than the maximum concurrency, then there are spare channels.  
-   7. If there are no reserved channels available for any system, and there are spare channels, it tries to schedule a task based on probability.
-2. For progressing a task based on probability, it iterates over all the systems in the configuration by systemName
-   1. It finds the number of free channels available and calculates the probability range for the system
+It carries out the operations as follows:
+
+1. The Lambda makes a call to SSM to get the flow control config. It adds it to a class object which validates the config for:
```
MancunianSam: Can you put "SSM parameter store" please, there's dozens of SSM features.

MancunianSam (commented):
Looks good thanks, just a couple of minor bits. Can you add a unit test for the writer a bit like this https://github.com/nationalarchives/dr2-ingest/blob/main/scala/lambdas/dynamo-formatters/src/test/scala/uk/gov/nationalarchives/dynamoformatters/DynamoFormattersTest.scala#L630 please

MancunianSam (approved):

